PR #7, opened 2017-01-08
Added MD2 hash function

felipeamp:
I added the MD2 hash function and 3 basic tests (the ones on the Wikipedia page). I based the structure on the MD4 hash, but I'm not sure if the example code is correct.

If there are any problems, please let me know. :)


md2/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "md2"
+version = "0.4.0"
```
newpavlov: It's better to start from 0.1.0, as it's not required to synchronize versions across crates.

md2/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+documentation = "https://docs.rs/md2"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["crypto", "md2", "hash", "digest"]
+
+[dependencies]
+digest = "0.4"
+digest-buffer = "0.2"
+generic-array = "0.6"
+byte-tools = "0.1"
+fake-simd = "0.1"
```
newpavlov: You don't use neither byte-tools, nor fake-simd, so you can remove them from here.

md2/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2016 bacher09, Artyom Pavlov
```
newpavlov: It's your work, so you can insert your own credits instead of mine and bacher09. ;)

md2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
+//! The [MD2][1] hash function.
+//!
+//! [1]: https://en.wikipedia.org/wiki/MD2_(cryptography)
+
+#![no_std]
+extern crate generic_array;
+extern crate byte_tools;
```
newpavlov, line 6: Not used byte_tools.

md2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+    fn new() -> Md2State {
+        Md2State {
+            x: Default::default(),
+            checksum: Default::default(),
+        }
+    }
+
+    fn process_block(&mut self, input: &Block) {
+        // Update state
+        for j in 0..16usize {
```
newpavlov: `usize` is not necessary here. Rust can infer type in this situation.

md2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            self.x[16 + j] = input[j];
+            self.x[32 + j] = self.x[16 + j] ^ self.x[j];
+        }
+
+        let mut t = 0u8;
+        for j in 0..18u8 {
+            for k in 0..48usize {
```
newpavlov: Same for lines 64-65. (you can remove `u8` from from 63 too, but I guess in that case it's better to stay explicit)

md2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            // Padding
+            let rem = self.buffer.remaining();
+            let mut buffer_end = self.buffer.next(rem);
+            for idx in 0..rem {
+                buffer_end[idx] = rem as u8;
+            }
+        }
+        self_state.process_block(self.buffer.full_buffer());
+
+        let checksum = self_state.checksum.clone();
```
newpavlov: You don't need `clone` here.

md2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+        self.buffer.input(input, |d: &Block| {
+            self_state.process_block(d);
+        });
+    }
+
+    fn result(mut self) -> GenericArray<u8, Self::OutputSize> {
+        self.finalize();
+
+        let mut out = GenericArray::default();
+        for (x, y) in self.state.x[0..16].iter().zip(out.iter_mut()) {
```
newpavlov: The simple loop:
```
for i in 0..16 {
    out[i] = self.state.x[i];
}
```
will be more readable and should have the same efficiency. (compiler can remove bound checks in this case) Or you can use `copy_memory` from `byte_tools`.

md2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            }
+        }
+        self_state.process_block(self.buffer.full_buffer());
+
+        let checksum = self_state.checksum.clone();
+        self_state.process_block(&checksum);
+    }
+}
+
+impl Default for Md2 {
```
newpavlov: Actually instead of implementing `Default` like this you can simply derive it for `Md2` and `Md2State`. After that you will not need `new()` for `Md2State` and you will be able to simply use `Default::default()` in `Md2::new()`. (well some argue that there is no need for `new` without any arguments and that `Default` trait would be sufficient, maybe in the future I will remove `new()` altogether)

md2/examples/md2sum.rs
```diff
@@ -0,0 +1,49 @@
...
+        let n = match reader.read(&mut buffer) {
+            Ok(n) => n,
+            Err(_) => return,
+        };
+        sh.input(&buffer[..n]);
+        if n == 0 || n < BUFFER_SIZE {
+            break;
+        }
+    }
+    print_result(&sh.result(), name);
```
newpavlov: [Now](https://github.com/fizyk20/generic-array/blob/master/src/hex.rs) we can write simply `println!("{:x}\t{}", sh.result(), name)`, so `print_result` function is not necessary.

md2/src/lib.rs
```diff
@@ -0,0 +1,132 @@
...
+            for k in 0..48usize {
+                self.x[k] ^= S[t as usize];
+                t = self.x[k];
+            }
+            t = t.wrapping_add(j);
+        }
+
+        // Update checksum
+        let mut l = self.checksum[15];
+        for j in 0..16usize {
```
newpavlov: Same for `usize`.

md2/tests/lib.rs
```diff
@@ -0,0 +1,19 @@
...
+
+#[test]
+fn md2_main() {
+    // Examples from wikipedia
+    let tests = new_tests!("test1", "test2", "test3");
+    main_test::<md2::Md2>(&tests);
+}
+
+// #[test]
+// fn md2_1million_a() {
```
newpavlov: I guess it's better to add 1million_a test, e.g. by generating it by another tool, or remove it altogether.

newpavlov (changes requested):
Thank you for you work!
Some minor changes and we are good to go!

md2/src/lib.rs
felipeamp: Actually, since checksum is inside the Md2State struct and is is already mutably borrowed through self_state, the compiler complains about borrowing checksum. Should I still clone it or should I move the checksum to the Md2 struct?

md2/src/lib.rs
felipeamp: Totally agree!

md2/tests/lib.rs
felipeamp: At this point I guess I'll just remove it. Later I'll try to create the 1million_a test file.

md2/examples/md2sum.rs
felipeamp: I tried doing that, but the compiler complains that `<D as md2::Digest>::OutputSize` (which is `generic_array::typenum::U16`) does not implement `std::ops::Add`.
Any ideas?

newpavlov: Hm, implementation of `LowerHex` and `UpperHex` requires additional trait bounds to create array with double length internally. The easiest way will be to remove unnecessary generalization over `Digest` and simply create `Md2` instance inside `process`:
```
fn process<R: Read>(reader: &mut R, name: &str) {
    let mut sh = Md2::new();
    ...
    println!("{:x}\t{}", &sh.result(), name);
}
```

md2/src/lib.rs
newpavlov: This should work:
```
let checksum = self_state.checksum;
self_state.process_block(&checksum);
```
But this will not:
```
self_state.process_block(&self_state.checksum);
```
There is still issues with borrow checker. If I understand situation correctly in the first case `Copy` trait works behind the scene as with `self_state` at the first line of the method.

felipeamp: Aha. I would never have thought that would work while the other doesn't.

felipeamp:
Anything else?

newpavlov (approved):

newpavlov:
Everything looks good! Thank you!

PR #9, opened 2017-01-11
Added the one_million_a test to MD2 Hash

felipeamp:
Should have done this before. Better late than never!


newpavlov:
Thank you again!

PR #17, opened 2017-03-03
Blake2 rework

newpavlov:
Rework of blake2 crate based on [`blake2-rfc`](https://github.com/cesarb/blake2-rfc). (with the consent of Cesar)


Ralith:
I relied on 0.3.0's support for statically customized hash output size (specifically, I use a 200-bit hash because it divides evenly into both bytes and base 36). This seems to break my use case.

newpavlov:
BLAKE2 defines variable size output as a simple truncation, so this code from compiler perspective will behave identically to the old approach:

```Rust
use blake2::Blake2b;
use digest::{Digest, VariableOutput};

let hasher = Blake2b::new();
// do hashing
let hash = [0u8; 200];
hasher.variable_result(&mut hash[..]);
```

Ralith:
Thanks! However, doesn't [the 0.3.0 code](https://github.com/RustCrypto/hashes/blob/ba3383dd404ef8dee0cf8388ed4ef59ea4fb615b/blake2/src/blake2b.rs#L112) place the output size in the parameter block, as specified in [the RFC](https://tools.ietf.org/html/rfc7693#section-2.5)? It seems like the new approach would be non-conformant and yield different hashes.

newpavlov:
Oh... It indeed looks like my mistake and overlook while reworking the crate. Thank you for noticing this!

I will try fix this issue until the end of the week.

Ralith:
Thanks! No particular urgency, since 0.3.0 is still working fine for me and I can always construct a parameter block manually if I *really* need to, but it's good to know support is intended.

PR #32, opened 2017-07-17
Implementation of ISO crc32

althonos:
Hi there,
I took the time to implement the CRC32 *checksum* using a lookup table, so that performance is actually quite good. Unfortunately, the `crc32` package name is taken on crates.io, so I went with `crc-32`.

The values in the tests were computed with the `cksfv` command line tool, and cross validated with some [online CRC32 calculator](https://www.lammertbies.nl/comm/info/crc-calculation.html).

The implemented CRC32 is the *standard* one (although there is no such thing), used in the Ethernet protocol, Zlib, etc. I could probably implement the CRC-32C (Castagnoli) if you're interested.


tarcieri:
I'm not sure it makes sense to include CRC32 in a cryptography library

althonos:
Well, this library includes `md5`, which is not cryptographically secure but widespread for checksums. So it is safe to assume softwares depending on the `md-5` crate are going to use it as a checksum, not a a secure hash. It then makes sense to have an open-source CRC32 implementation as well.

gsingh93:
I also don't think this belongs here.

I think this repo is for cryptographic hashes, even if those cryptographic hashes have known vulnerabilities.

CRC32 just isn't even a cryptographic hash function.

tarcieri:
Yes, I helped add the "security level" indicators to this project, which note MD5 (and now SHA1) are broken.

While they are broken, they were at least designed to be cryptographic hash functions. CRC32 is obviously not.

newpavlov:
I agree with others, there is a difference between cryptographic hash function with vulnerabilities and hash functions not designed as cryptographic ones. Also I am not sure if it's a good decision to use traits from `digest` crate for CRC32, as its description states that traits in it used for cryptographic hash functions, so it can create some confusion.

I wrote to @sivadeilra (owner of `crc32` crate) about merging your work into his crate or about possible transfer of crate ownership to you. Hope he will answer.

althonos:
Okay, no problem !
checksum

I just thought it would fit the crate, but actually I get your point. I'll be closing the PR, sorry for the disturbance ! 😄 
______________
**EDIT:** The crate is called `pruefung`, and implements various legacy checksums in pure Rust (see repository [here](https://github.com/althonos/pruefung))

PR #39, opened 2017-10-16
Minor updates to codestyle

kylegalloway:
Hi,
I've updated much of your code based on clippy recommendations. I've tried to keep speed and readability where possible so there are some places I didn't follow clippy's recommendations.


gost94/src/gost94.rs
```diff
@@ -15,10 +15,10 @@ pub type Block = GenericArray<u8, U32>;
 
 fn sbox(a: u32, s: &SBox) -> u32 {
     let mut v = 0;
-    for i in 0..8 {
+    for (i, item) in s.iter().enumerate().take(8) {
```
newpavlov: I believe explicit loop over `i` conveys intent much better in such cases and has better readability, so I think it's better to reverse such changes.

groestl/src/state.rs
```diff
@@ -151,11 +151,8 @@ impl<BlockSize> GroestlState<BlockSize>
         for i in 0..matrix.rows() {
             for j in 0..matrix.cols() {
                 matrix[i][j] ^= c[i * 16 + j];
-                if c[0] == 0x00 && i == 0 {
-                    matrix[i][j] ^= round;
-                } else if c[0] == 0xff && i == 7 {
-                    matrix[i][j] ^= round;
-                }
```
newpavlov: This change is not equivalent and I think it's the reason behind failing tests.

md4/src/lib.rs
```diff
@@ -38,30 +38,30 @@ pub struct Md4 {
 
 impl Md4State {
     fn process_block(&mut self, input: &Block) {
-        fn f(x: u32, y: u32, z: u32) -> u32 {
+        fn foo(x: u32, y: u32, z: u32) -> u32 {
```
newpavlov: IIRC `f`, `g` and `h` are names used in the description of the algorithm, so it's much better to use them instead of `foobar`s.

streebog/src/table.rs
```diff
@@ -3,6 +3,8 @@
 //!
 //! It was created using `gen_table` function
 
+#[cfg_attr(feature = "cargo-clippy", allow(unreadable_literal))]
```
newpavlov: Not sure why you allow `unreadable_literal` here and in whirlpool, but in other places add underscores.

blake2/src/consts.rs
```diff
@@ -14,10 +14,10 @@ pub static SIGMA : [[usize; 16]; 12] = [
 ];
 
 pub static BLAKE2B_IV : [u64; 8] = [
-    0x6a09e667f3bcc908, 0xbb67ae8584caa73b,
-    0x3c6ef372fe94f82b, 0xa54ff53a5f1d36f1,
-    0x510e527fade682d1, 0x9b05688c2b3e6c1f,
-    0x1f83d9abfb41bd6b, 0x5be0cd19137e2179,
```
newpavlov: Constants are usually taken directly from the specifications and not intended to be read by humans, so I am not sure if its worth to introduce such change. @tarcieri what do you think?

newpavlov (commented):
Thank you for your effort!

But I am not sure if we should use some of the changes in this PR. I hope @tarcieri will add his input to my comments.

blake2/src/consts.rs
kylegalloway: I didn't realize this was the case until later on when I added the explicit allow for unreadable_literals in constant files. I can go back and do that here.

gost94/src/gost94.rs
kylegalloway: No problem

groestl/src/state.rs
kylegalloway: Yes, For some reason I read that as:
```
if a {
    do_something;
} else {
    do_something;
}
```

It should actually be:
```
if a || b {
    do_something;
}
```

md4/src/lib.rs
kylegalloway: No problem.

streebog/src/table.rs
kylegalloway: This is what I should do in most other places.

kylegalloway:
Should I make the changes on this PR or should I make all of the changes in a separate PR to make it look cleaner?

kylegalloway:
Alright. It is now passing all tests and has many of the clippy lints fixed. Please review it at your leisure.

blake2/src/consts.rs
```diff
@@ -13,6 +13,7 @@ pub static SIGMA : [[usize; 16]; 12] = [
     [ 14, 10,  4,  8,  9, 15, 13,  6,  1, 12,  0,  2, 11,  7,  5,  3 ],
 ];
 
+#[cfg_attr(feature = "cargo-clippy", allow(unreadable_literal))]
```
newpavlov: Can't we use `#![cfg_attr(..))]` on the whole file? I think it will be less repetitive. Same for other `consts.rs` files.

sha1/src/consts.rs
```diff
@@ -4,14 +4,14 @@ pub const STATE_LEN: usize = 5;
 pub const BLOCK_LEN: usize = 16;
 
 #[cfg(not(feature = "asm"))]
-pub const K0: u32 = 0x5A827999u32;
+pub const K0: u32 = 0x5A82_7999u32;
```
newpavlov: I think we should `allow(unreadable_literal)` here too for consistency.

newpavlov (commented):
Two small comments and we good to merge! You can leave this PR as is, I will use squash and merge, so history of commits will stay clean.

kylegalloway:
Ok, I fixed a the concerns you had and added some module level lint allows so that if you run clippy without arguments it will pass with flying colors! I also double checked that the build/tests pass. Please review it at your leisure.

tarcieri:
Looks good to me

newpavlov:
Thank you! I will merge your PR.

P.S.: One small thing which I've missed is that `u16::from(1)` is better written as `1u16`, but I'll fix it myself.
EDIT: Ups, it was `l`, not `1`. :)

newpavlov:
I will publish your updates with the update on next minor version of `digest` crate, which I plan to do in the near future.

Issue #37, opened 2017-08-13
CI improvements


newpavlov:
- [x] Add crate specific feature tests (including `asm`)
- [x] Add big-endian tests (using either qemu or https://github.com/japaric/trust.) [#40]
- [x] Test minimal Rust versions for crates (currently all crates support 1.13 or higher)
- [x] Add tests to check if crates are truly `no_std` (see #36)

newpavlov:
The oldest Rust version which is continuously tested will be 1.18 to simplify CI config by using only `test --all`

quininer:
We can use [utest](https://github.com/japaric/utest) to do `no-std` test. (thanks @japaric !!)

~~`digest` can not currently be build because `generic-array` specifies a [`typenum` with std](https://github.com/fizyk20/generic-array/blob/master/Cargo.toml#L15).~~

```
$ xargo build
    Updating registry `https://github.com/rust-lang/crates.io-index`
   Compiling byte-tools v0.2.0
   Compiling arrayref v0.3.4
   Compiling typenum v1.9.0
error[E0463]: can't find crate for `std`

error: aborting due to previous error

error: Could not compile `typenum`.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

newpavlov:
`typenum` always [enables](https://github.com/paholg/typenum/blob/master/src/lib.rs#L46) `#![no_std]`. Problem originates from `typenum` building process which [requires](https://github.com/paholg/typenum/blob/8ff62fafe1a8127638a307d5ead80ec1bf9794c0/build/main.rs) `std`. I think xargo should be able to handle such cases correctly.

quininer:
You are right.

japaric:
@quininer Try with `xargo build --target x86_64-unknown-linux-gnu`. `xargo build` will build the build scripts against the Xargo sysroot, which doesn't have std.

newpavlov:
I've added `xargo build` to CI config, I think it should be enough to catch slipped `std` in future. Thank you @japaric!

PR #40, opened 2017-10-31
Add big-endian ci (powerpc)

quininer:
This PR will resolve part of #37.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,24 @@
 language: rust
-rust:
-  - stable
-  - beta
-  - nightly
-matrix:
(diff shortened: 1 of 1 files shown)


newpavlov:
Thank you! It looks like we indeed have some crates (gost94 and groestl at least) bugged on BE architecture.

quininer:
I think we should avoid using usize.

newpavlov:
Yes, I will skim through code and see if we can remove some of `usize`s, but if I remember correctly almost all of the uses should be related to indexing over tables or buffers.

PR #46, opened 2017-12-01
add blake

kazcw:
I wrote a rust implementation of the "Blake" SHA3 finalist, based closely on the reference implementation. It's a little raw; it needs docs, lots more tests, and a macro cleanup I think. But it's getting the right results (I've tested it on a lot more random inputs than are currently committed to repo), it performs fine, and I think I've implemented the same API as the other crates in this workspace.


newpavlov:
Thank you for your contributions!

Unfortunately it will be some time before I will be able to check your PRs. I'll try to do it in December!

kazcw:
Thank you for all the work you're doing on this project! Unifying around thoughtfully-designed APIs is so important for building up a pure Rust crypto ecosystem.

kazcw:
I no longer have time to keep synchronizing this PR with the published crate. BLAKE can be found here: https://github.com/cryptocorrosion/cryptocorrosion/tree/master/hashes/blake

PR #47, opened 2017-12-02
add skein

kazcw:
Depends on RustCrypto/block-ciphers#14 and RustCrypto/utils#5.


kazcw:
I no longer have time to keep synchronizing this PR with the published crate. Skein can be found here: https://github.com/cryptocorrosion/cryptocorrosion/tree/master/hashes/skein

PR #63, opened 2018-10-20
keccak1600

kazcw:
Add a sha3 variant needed for CryptoNight.


newpavlov:
Sorry for the late reply!

Looks good to me, but I would like to add tests for this varaint. Do you know if CryptoNight has published test vectors?

kazcw:
Monero doesn't have unit tests for this function, but the full CryptoNight hash is keccak1600 chained with another hash function; so I could derive tests from the CryptoNight test vectors that will known to be correct as long as there isn't a collision of the following 32-bit hash function that happens to make wrong keccak outputs produce the right final output.

newpavlov:
>I could derive tests from the CryptoNight test vectors

It would be great if you'll do it. You can just past them here in hex representation and I will do conversion into the binary format used by test macros.

kazcw:
I made a special branch of my CryptoNight crate that has the inputs and outputs in hex. It tests keccak1600 outputs, and also that those keccak outputs yield the correct final hashes from Monero's test vectors. There are 4 cases based on CryptoNight v0, and 8 based on CryptoNight v2.
Source is here: https://github.com/kazcw/yellowsun/blob/test-keccak/src/lib.rs#L171

newpavlov:
Thanks! I will port them to the binary format and add to this PR.

Issue #67, opened 2018-12-13
Add `Blake2b-256`, `Blake2b-128`, etc. as wrappers over `VarBlake2b`.


dpc:
It looks like some crates want a `D: Digest<OutputSize = U64> + Default` as an argument to what hash to use, so that `VarBlake2b` is not very useful on it's own.  Eg. https://github.com/dalek-cryptography/ed25519-dalek/issues/59

I could write my own wrapper, but if I need it, other people will, too.

newpavlov:
Yes, we should do it as part of `blake2`. Though instead of wrappers a better option will be to write a generic implementation over output size (something like `GenBlake2b<N>`), maybe with some convenience aliases (e.g. `type Blake2b256 = GenBlake2b<U32>`). Luckily `typenum` allows us to [define](https://docs.rs/typenum/1.10.0/typenum/operator_aliases/type.LeEq.html) "less or equal" bounds, though IIRC we will not be able to migrate it straightforwardly to const generics in future. I think older versions of `blake2` had such genericity, but were removed in favor of implementing `VariableOutput`.

newpavlov:
Forgot to ask: can you do a PR with these updates?

dpc:
Most people will want 128, 256, 512 versions only, and they can be always changed to whatever is the best option for the internal type, in the future, no? In a backward compat way, for whichever method is used. So is there any point to complicate everything with `typenum` already?

dpc:
I could do it with a simple macro... :D

newpavlov:
Crates already have been "complicated" with `typenum`. :) I would prefer to have a more universal and flexible solution (e.g. how about 384 bit hashes used with signatures? or maybe someone will want to replace CRC32 checksum?), plus it will not be that difficult to implement, you'll just need to make generic this [struct](https://github.com/RustCrypto/hashes/blob/master/blake2/src/blake2.rs#L282) and associated impls, and create necessary type aliases.

dpc:
Ok then I'll see what  I can do.

dpc:
I would like to confirm something? Is the only difference between Blake with 512bit output and Blake with 256bit output, that the shorter one does everything the same and takes only first 32 bytes of the result? Is this standardized somewhere, arbitrary, common convention? I'm not a cryptographer.

burdges:
[...]

PR #69, opened 2018-12-15
Add fixed-output-size blake2 algorithms

dpc:
Attempt at #67

### DIFF
--- a/blake2/src/blake2.rs
+++ b/blake2/src/blake2.rs
@@ -15,6 +15,8 @@ macro_rules! blake2_impl {
         use core::cmp;
         use byte_tools::{copy, zero};
         use crypto_mac::{Mac, MacResult, InvalidKeyLength};
+        use std::marker::PhantomData;
+        use digest::generic_array::ArrayLength;
 
         type Output = GenericArray<u8, $bytes>;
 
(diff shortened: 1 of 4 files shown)


blake2/src/blake2.rs
```diff
@@ -279,56 +281,61 @@ macro_rules! blake2_impl {
...
-        impl FixedOutput for $fix_state {
-            type OutputSize = $bytes;
+        impl<U: ArrayLength<u8>> FixedOutput for $fix_state<U> {
+            type OutputSize = U;
 
-            fn fixed_result(self) -> Output {
-                self.state.finalize_with_flag(0)
+            fn fixed_result(self) -> GenericArray<u8, Self::OutputSize> {
+
+                let state_out = self.state.finalize_with_flag(0);
```
dpc, line 311: Is this OK?

dpc:
TODO: Add  "less or equal" bounds?

blake2/src/blake2.rs
dpc: I don't think it is. I should somehow use the `variable_result`, but do I still need to call `finalize_with_flag`?

blake2/src/blake2.rs
```diff
@@ -279,56 +281,61 @@ macro_rules! blake2_impl {
...
-                let state = $state::new_keyed(&[], $bytes::to_usize());
-                Self { state }
+                let state = $state::new_keyed(&[], U::to_usize());
+                Self { state, _u: PhantomData }
             }
         }
 
-        impl BlockInput for $fix_state {
+        impl<U> BlockInput for $fix_state<U> {
             type BlockSize = $bytes;
```
dpc, line 297: It seems (from wikipedia) that block size should be 32 for Blake2b256. But what about intermediate values?

newpavlov: Changing output size does not change the block size.

blake2/src/blake2.rs
newpavlov: Looks good to me, you don't have to reuse `variable_result`, especially considering that it's a wrapper around fixed output version.

blake2/src/blake2b.rs
```diff
@@ -1,8 +1,17 @@
...
 );
+
+/// Blake2b with 128 bit output
+pub type Blake2b128 = FixedBlake2b<U16>;
+/// Blake2b with 256 bit output
+pub type Blake2b256 = FixedBlake2b<U32>;
+/// Blake2b with 512 bit output
+pub type Blake2b512 = FixedBlake2b<U64>;
+/// Blake2b with 512 bit output
+pub type Blake2b = FixedBlake2b<U64>;
```
newpavlov, line 17: Because you have used `struct $fix_state<U = $bytes> { .. }`, you don't have to introduce `FixedBlake2b`. You can write `type Blake2b256 = Blake2b<U32>;` and remove `Blake2b512` as a redundant one. Same goes for `Blake2s`

blake2/src/blake2.rs
newpavlov: Ah, it's the other way around with wrappers, but still I think it's OK to call `finalize_with_flag` here. Plus you can write this part like this:
```rust
let state_out = self.state.finalize_with_flag(0);
GenericArray::clone_from_slice(&state_out[..U::to_usize()])
```

newpavlov:
Overall looks good to me!

>Add "less or equal" bounds?

Yes, it would be nice. Code has asserts for this, but I think it's better to have compile-time checks if possible.

blake2/src/blake2b.rs
dpc: > Because you have used struct $fix_state<U = $bytes> 

This actually did not fully work when running test suite. For some reason `new()` could not take `U64` as the default. My guess was that it's because `new` is defined on a `impl SomeTrait fo $fix_state<U = $bytes>`r , and not directly.

newpavlov: Hm, I will need to look into this... This behavior may create issues in `blowfish` crate which switched to default type parameters recently.

dpc: I'll try to get back to it, but can't tell why exactly. Feel free to take over this PR if you'd like.

tarcieri:
I believe this PR is obsolete?

newpavlov:
I wouldn't say obsolete, I will try to revive it later.

Issue #31, opened 2017-07-03
Implementations of BLAKE2sp and BLAKE2bp


pedrocr:
Apparently BLAKE2 includes parallel versions of the two hash functions to take advantage of multi-core:

https://github.com/BLAKE2/BLAKE2/blob/master/sse/blake2sp.c

It would be nice to have these to be able to take advantage of more cores when hashing large files.

Ralith:
These would also be useful as SIMD implementations, enabling dramatic speedups on common hardware without any extra threads at all.

tomtau:
btw, here's a bounty on this issue: https://www.bountysource.com/issues/46854589-implementations-of-blake2sp-and-blake2bp

pczarn:
I started working on the issue.

I am currently making a script to convert test data from the `BLAKE2` repo to blobs in the `hashes` repo.

tarcieri:
FYI, I just rebased #228 which completely replaces the current implementation with `blake2-simd`

pczarn:
@tarcieri Okay, I can see that `blake2-simd` includes both parallel variants.

newpavlov:
Closing in favor of #1.

PR #70, opened 2019-01-23
blake2bp/blake2sp

kazcw:
No-std (single-threaded) implementation.

Tests included are the official test vectors plus a long-input case.

cf. #31


newpavlov:
Thank you! Can you please fix the CI failures? Looks like you are not handling big-endian correctly and use absolute path which do not work in Rust 1.21.

kazcw:
Fixed the bugs; also added bp/sp benchmarks.

tarcieri:
@kazcw are you still interested in landing this and if so, can you rebase please?

kazcw:
@tarcieri Upstream has seen enough change in the last year and a half that I think it would be easier to start over than reconcile this.

Issue #56, opened 2018-06-21
Port blake2 on coresimd


newpavlov:
With stabilization of SIMD intrinsics in Rust 1.27 we can remove `simd` feature from the `blake2` crate.

tomtau:
btw, here's a bounty on this issue: https://www.bountysource.com/issues/59948303-port-blake2-on-coresimd

kazcw:
How would this work for rustc 1.21..=1.26?

newpavlov:
For v0.8 it will have to be feature-gated behind some feature (it can be the same `simd` feature), but I am thinking about migrating to Rust 2018 for v0.9.

DonaldTsang:
What about BLAKE3?

tarcieri:
I’d be very much into us having a high-performance BLAKE3 implementation but am not aware of the state-of-the-art in Rust BLAKE3 libraries /cc @zooko

tarcieri:
See BLAKE3 #103

pczarn:
I started working on the issue.

tarcieri:
@pczarn it'd be really good to get #228 landed before we try to extend the current implementation

pczarn:
@tarcieri Okay, so you are close to getting a new implementation, but a migration from `core::arch` x86/x86_64 intrinsics to a portable `core::simd` will remain to be done?

pczarn:
Looks like `coresimd` is some internal utility, and not `core::simd`, so:

* could you point me to the code for `coresimd`?
* would it be better to give it a name that cannot be confused with core's `simd`?
* are you looking into using core's `simd` once it gets stabilized?

tarcieri:
I think this ticket is actually referring to `core::arch` SIMD intrinsics, though adding support for `core::simd` would be nice

pczarn:
Ok, please update me when progress with #228 happens

PR #72, opened 2019-01-25
blake2: x86 coresimd

kazcw:
cf. #56

Replacement for simd + simd_opt features. Builds on stable rustc 1.27+

The coresimd feature overrides all other simd flags. Build with coresimd will fail for unsupported target platforms, e.g. non-x86.

Benched on a Xeon X5650:
sse2 only: faster than simd_opt
ssse3: slightly faster than simd_opt

Benched on a Haswell:
avx2: same speed as simd_opt

NB. the standard CI settings won't test this properly since it introduces a new feature


newpavlov:
Thank you! 

You can introduce a separate line for CI like was [done](https://github.com/RustCrypto/block-ciphers/blob/master/.travis.yml#L14) for testing AES-NI. I will try to review both of your blake2 PRs on this weekend.

tarcieri:
@kazcw are you interested in continuing to work on this PR, and if so, can you rebase?

kazcw:
@tarcieri As with #70, better to start from scratch at this point.

PR #74, opened 2019-01-29
JH

kazcw:
Import of the jh-x86_64 crate. Portable implementation plus compile-time-selected coresimd optimizations for sse2, ssse3, and avx2. This is, as far as I'm aware, the fastest implementation of JH due to the ssse3 optimizations, which improve performance to about 105.5% of the NIST submission's SSE2-bitsliced implementation on my machine.

Because this crate uses coresimd, merge will either need to wait until this project's minimum supported Rust version is increased, or I could feature-gate the optimizations. The portable fallback runs at a little under half the speed, and is about 10% faster than the "optimized 64-bit" implementation from the NIST package on my Nehalem box.

cf. #1


kazcw:
I went ahead and made it Rust 1.21-compatible, with simd an opt-in feature

kazcw:
The build failures are due to block-buffer and hex-literal, but I'm not sure why

kazcw:
This PR is getting out of sync. The crate can be found here: https://github.com/cryptocorrosion/cryptocorrosion/tree/master/hashes/jh

PR #78, opened 2019-04-11
Add support for Persona and Salt to Blake2

### LINKED ISSUE #75: Add support for Persona and Salt to Blake2
I would like to add these, based on this PR: https://github.com/cesarb/blake2-rfc/pull/9 but the current traits for `Digest` don't really allow for something that has custom parameters. 

@newpavlov  My question is, should something like `with_params` be added to an existing trait, should this be a new trait, be just a one off thing for Blake, or am I missing something.

str4d:
Closes #75.

### DIFF
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -12,7 +12,7 @@ categories = ["cryptography", "no-std"]
 [dependencies]
 digest = "0.8"
 byte-tools = "0.3"
-byteorder = "1"
+byteorder = { version = "1", default-features = false }
 crypto-mac = "0.7"
 opaque-debug = "0.2"
 
@@ -23,7 +23,7 @@ hex-literal = "0.1"
 
 [features]
 default = ["std"]
-std = ["digest/std", "crypto-mac/std"]
+std = ["digest/std", "byteorder/std", "crypto-mac/std"]
 simd = []
 simd_opt = ["simd"]
 simd_asm = ["simd_opt"]
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -12,6 +12,7 @@ categories = ["cryptography", "no-std"]
 [dependencies]
 digest = "0.8"
 byte-tools = "0.3"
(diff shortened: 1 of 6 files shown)


str4d:
This PR is ported from https://github.com/cesarb/blake2-rfc/pull/9

burdges:
I'd think the byteorder crate is kinda deprecated by `std::u32/u64::to_le_bytes/from_le_bytes`, etc. now.

str4d:
Neat, I hadn't spotted those! However, using them would raise the minimum-supported Rust version to 1.32, while it is currently 1.21.

str4d:
@newpavlov do you have any feedback on this PR? Based on https://github.com/RustCrypto/hashes/issues/75#issuecomment-466530062 I assumed that you'd be happy with a direct port of the `blake2-rfc` PR.

str4d:
To anyone else looking for personalisation, the [blake2b_simd](https://crates.io/crates/blake2b_simd) and [blake2s_simd](https://crates.io/crates/blake2b_simd) crates have personalisation APIs. They also mandate the use of SIMD when available however, so I think it's still valuable to have this PR merged for those projects that don't want a SIMD dependency.

blake2/Cargo.toml
```diff
@@ -12,6 +12,7 @@ categories = ["cryptography", "no-std"]
 [dependencies]
 digest = "0.8"
 byte-tools = "0.3"
+byteorder = { version = "1", default-features = false }
```
tarcieri, line 15: Curious if it might make sense to start requiring Rust 1.32+ and using the built-in endianness conversions rather than the `byteorder` crate

tarcieri:
@str4d still interested in this, and if so, mind rebasing?

str4d:
Rebased on current master.

blake2/Cargo.toml
tarcieri: I can circle back on this after we bump the `digest` crate and with it MSRV

PR #81, opened 2019-04-22
Turn Keccak256Full comment into docs

felixrabe:
/// Docs...
    item

is equivalent to:

    #[doc="Docs..."]
    item

See https://stackoverflow.com/a/33999625/1034080 for an example.

### DIFF
--- a/sha3/src/lib.rs
+++ b/sha3/src/lib.rs
@@ -68,17 +68,17 @@ mod state;
 pub use reader::Sha3XofReader;
 use state::Sha3State;
 
-sha3_impl!(Keccak224, U28, U144, paddings::Keccak);
-sha3_impl!(Keccak256, U32, U136, paddings::Keccak);
-sha3_impl!(Keccak384, U48, U104, paddings::Keccak);
-sha3_impl!(Keccak512, U64, U72, paddings::Keccak);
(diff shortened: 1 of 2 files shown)


newpavlov:
Looks good! I will appreciate if you also will make `meta` argument a required one and will add docs for other types.

felixrabe:
- [x] Make `meta` argument required.
- [ ] Add docs for other types.

I'll have a look.

newpavlov:
BTW you also can pass docstrings only, see for example `vardoc` and `doc` macro argumetns in [BLAKE2](https://github.com/RustCrypto/hashes/blob/master/blake2/src/blake2.rs).

felixrabe:
Ok, changed the macro argument from `$meta:meta` to `$doc:expr` and put it to the end, so the macro invocations still more or less line up.

About the empty docs for the other types: I don't really know SHA-3 and Keccak well enough, nor do I have the energy to research them ATM (I got stuck yesterday reading half of HN about pros and cons of SHA-2 and SHA-3).

Do you have any other input for me about the proposed changes here? Otherwise I suggest you merge it and fill in the blanks yourself.

newpavlov:
No problem, I will add docs a bit later! Thank you!

PR #82, opened 2019-04-29
impl `Digest` for ring

flier:
I know `ring` has it's [Context](https://briansmith.org/rustdoc/ring/digest/struct.Context.html) trait, but [digest::Digest](https://docs.rs/digest/0.8.0/digest/trait.Digest.html) seems be better to integrate other digest implementation, such as `openssl-digest` later.


ring-digest/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "ring-digest"
+version = "0.8.0"
+authors = ["RustCrypto Developers"]
+license = "MIT OR Apache-2.0"
+description = "RIPEMD-160 hash function"
+documentation = "https://docs.rs/ring-digest"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["crypto", "ring", "hash", "digest", "sha1", "sha2", "SHA-1", "SHA-256", "SHA-384", "SHA-512", "SHA-512/256"]
```
newpavlov: I think `SHA-*` keywords are redundant (none of the other crates use them), plus IIRC crates.io limits number of keywords to 5.

ring-digest/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+categories = ["cryptography", "no-std"]
+
+[dependencies]
+digest = { version = "0.8", features = ["std"] }
+opaque-debug = "0.2"
+ring = "0.14"
```
newpavlov: Adding `default-features = false` should help with the `no_std` CI test.

ring-digest/LICENSE-MIT
```diff
@@ -0,0 +1,27 @@
+Copyright (c) 2006-2009 Graydon Hoare
+Copyright (c) 2009-2013 Mozilla Foundation
+Copyright (c) 2016 Artyom Pavlov
```
newpavlov: You can remove these 3 lines and replace them with:
```
Copyright (c) 2019 The RustCrypto Project Developers
Copyright (c) 2019 <your name>
```

newpavlov (commented):
I have some minor change requests, but otherwise your PR looks good! I guess we will have to use a separate CI test script for `ring-digest`. Do you know if it's possible to cross compile `ring` to PowerPC (we use it to check if code works correctly on big-endian targets)?

Also to be safe I think it will be nice to add test vectors for implemented function. You can simply copy `blb` files from other crates and copy-paste tests.

@briansmith any comments?

tarcieri:
> Do you know if it's possible to cross compile `ring` to PowerPC (we use it to check if code works correctly on big-endian targets)?

*ring* supports `power64` targets

flier:
> Also to be safe I think it will be nice to add test vectors for implemented function. You can simply copy blb files from other crates and copy-paste tests.

Thanks @newpavlov, I have added those test cases, anything else need to do for CI?

For `power64` target, it seems `ring` build.rs crashed, maybe we need push a PR to upstream later.

flier:
There is a new PR briansmith/ring#819 from @q66 will add `powerpc` supports for `ring`

q66:
I'm still not sure if it will be possible to add 32-bit powerpc support, as the `ecp_nistz256` implementation in OpenSSL requires assembly code and there is none for 32-bit ppc (looks like for openssl on 32-bit ppc, it's simply disabled) and `ring` seems to hard depend on it; I'm not experienced enough with either crypto nor ppc asm to do that myself. 64-bit ppc is possible/works with my PR though.

tarcieri:
You almost certainly don't want to be doing ECC in software on PPC32. Most PPC32s have non-constant-time arithmetic operations (e.g. multiply is faster for 0 and 1) which in conjunction with their generally low clock speed makes for a particularly noisy timing sidechannel.

newpavlov:
@flier
Can you please resubmit this PR to the [ring-wrappers](https://github.com/RustCrypto/ring-wrappers/) repository?

PR #94, opened 2019-11-17
Small code improvements

rodoufu:
The build is failing due to a problem on `streebog.rs:149` but I haven't changed that.

### DIFF
--- a/streebog/src/streebog.rs
+++ b/streebog/src/streebog.rs
@@ -1,6 +1,6 @@
-#![cfg_attr(feature = "cargo-clippy", allow(needless_range_loop, inline_always))]
+#![cfg_attr(feature = "cargo-clippy", allow(clippy::needless_range_loop, clippy::inline_always))]
 use digest::{Input, BlockInput, FixedOutput, Reset};
-use digest::generic_array::typenum::{Unsigned, U64};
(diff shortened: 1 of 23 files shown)


streebog/src/streebog.rs
```diff
@@ -1,6 +1,6 @@
-#![cfg_attr(feature = "cargo-clippy", allow(needless_range_loop, inline_always))]
+#![cfg_attr(feature = "cargo-clippy", allow(clippy::needless_range_loop, clippy::inline_always))]
 use digest::{Input, BlockInput, FixedOutput, Reset};
-use digest::generic_array::typenum::{Unsigned, U64};
+use digest::generic_array::typenum::U64;
```
newpavlov: The build fails because you've removed `Unsigned` here, I am not sure why it happens, but until we bump MSRV lets keep it as-is.

rodoufu: You're right.
I've fixed it.

rodoufu:
@newpavlov at first I was making a change but some time after I realized that it wasn't necessary.
Then I ended up just removing some clippy warnings.

rodoufu:
@newpavlov do you still intend to review it?

tarcieri:
Thanks!

PR #96, opened 2020-01-03
Add AArch64 sha1 and sha2 extensions support

linkmauve:
Enables https://github.com/RustCrypto/asm-hashes/pull/10

`getauxval()` is extremely cheap, as it is just a pointer to the beginning of the stack, so it doesn’t make much sense to use a `lazy_static!()` to remember the flags (at least I can’t see any difference while profiling).


tarcieri:
> it doesn’t make much sense to use a lazy_static!()

Agreed, IMO `lazy_static` is effectively obsolete at this point

PR #97, opened 2020-01-04
Enable sha1 and sha2 AArch64 extensions from asm-hashes

linkmauve:
This follows https://github.com/RustCrypto/asm-hashes/pull/10

`getauxval()` is extremely cheap, as it is just a pointer to the beginning of the stack, so it doesn’t make much sense to use a `lazy_static!()` to remember the flags (at least I can’t see any difference while profiling).

Sorry about #96, I force-pushed from the wrong repository and can’t reopen the PR due to that…

Here are some benchmarks on my Nintendo Switch:

Without this patch (or without `--features=asm`):
```
test sha1_10      ... bench:          66 ns/iter (+/- 0) = 151 MB/s
test sha1_100     ... bench:         497 ns/iter (+/- 1) = 201 MB/s
test sha1_1000    ... bench:       4,799 ns/iter (+/- 16) = 208 MB/s
test sha1_10000   ... bench:      47,646 ns/iter (+/- 134) = 209 MB/s
test sha256_10    ... bench:          98 ns/iter (+/- 0) = 102 MB/s
test sha256_100   ... bench:         815 ns/iter (+/- 4) = 122 MB/s
test sha256_1000  ... bench:       7,980 ns/iter (+/- 76) = 125 MB/s
test sha256_10000 ... bench:      79,373 ns/iter (+/- 793) = 125 MB/s
```

With this patch:
```
test sha1_10      ... bench:          32 ns/iter (+/- 1) = 312 MB/s
test sha1_100     ... bench:         165 ns/iter (+/- 1) = 606 MB/s
test sha1_1000    ... bench:       1,480 ns/iter (+/- 5) = 675 MB/s
test sha1_10000   ... bench:      14,246 ns/iter (+/- 26) = 701 MB/s
test sha256_10    ... bench:          30 ns/iter (+/- 0) = 333 MB/s
test sha256_100   ... bench:         167 ns/iter (+/- 0) = 598 MB/s
test sha256_1000  ... bench:       1,554 ns/iter (+/- 7) = 643 MB/s
test sha256_10000 ... bench:      15,231 ns/iter (+/- 88) = 656 MB/s
```

### DIFF
--- a/sha2/src/aarch64.rs
+++ b/sha2/src/aarch64.rs
@@ -1,0 +1,9 @@
+// TODO: Import those from libc, see https://github.com/rust-lang/libc/pull/1638
+const AT_HWCAP: u64 = 16;
+const HWCAP_SHA2: u64 = 64;
+
+#[inline(always)]
+pub fn sha2_supported() -> bool {
+    let hwcaps: u64 = unsafe { ::libc::getauxval(AT_HWCAP) };
+    (hwcaps & HWCAP_SHA2) != 0
+}
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,18 @@ matrix:
     - env: NAME=features-test
       rust: nightly
       script: ./test_features.sh
+    - env: NAME=test-aarch64
+      arch: arm64
+      rust: nightly
+      script:
+        - cd sha1
+        - cargo test --verbose --release --features asm-aarch64
+        - cargo bench --verbose
+        - cargo bench --verbose --features asm-aarch64
+        - cd ../sha2
+        - cargo test --verbose --release --features asm-aarch64
+        - cargo bench --verbose
+        - cargo bench --verbose --features asm-aarch64
 
 install:
   - cargo install cross || true
--- a/sha1/Cargo.toml
+++ b/sha1/Cargo.toml
@@ -18,6 +18,7 @@ block-buffer = "0.7"
 fake-simd = "0.1"
 sha1-asm = { version = "0.4", optional = true }
 opaque-debug = "0.2"
+libc = { version = "0.2", optional = true }
 
 [dev-dependencies]
 digest = { version = "0.8", features = ["dev"] }
@@ -27,6 +28,10 @@ hex-literal = "0.1"
 default = ["std"]
 std = ["digest/std"]
 asm = ["sha1-asm"]
+
+# TODO: Remove this feature once is_aarch64_feature_detected!() is stabilised.
+# Only used on AArch64 Linux systems.
+asm-aarch64 = ["asm", "libc"]
 
 [badges]
 travis-ci = { repository = "RustCrypto/hashes" }
--- a/sha1/src/aarch64.rs
+++ b/sha1/src/aarch64.rs
@@ -1,0 +1,9 @@
+// TODO: Import those from libc, see https://github.com/rust-lang/libc/pull/1638
+const AT_HWCAP: u64 = 16;
+const HWCAP_SHA1: u64 = 32;
+
+#[inline(always)]
+pub fn sha1_supported() -> bool {
+    let hwcaps: u64 = unsafe { ::libc::getauxval(AT_HWCAP) };
+    (hwcaps & HWCAP_SHA1) != 0
+}
--- a/sha1/src/consts.rs
+++ b/sha1/src/consts.rs
@@ -2,16 +2,16 @@
 
 pub const STATE_LEN: usize = 5;
 
-#[cfg(not(feature = "asm"))]
+#[cfg(any(not(feature = "asm"), feature = "asm-aarch64"))]
 pub const BLOCK_LEN: usize = 16;
 
-#[cfg(not(feature = "asm"))]
+#[cfg(any(not(feature = "asm"), feature = "asm-aarch64"))]
 pub const K0: u32 = 0x5A827999u32;
-#[cfg(not(feature = "asm"))]
+#[cfg(any(not(feature = "asm"), feature = "asm-aarch64"))]
 pub const K1: u32 = 0x6ED9EBA1u32;
-#[cfg(not(feature = "asm"))]
+#[cfg(any(not(feature = "asm"), feature = "asm-aarch64"))]
 pub const K2: u32 = 0x8F1BBCDCu32;
-#[cfg(not(feature = "asm"))]
+#[cfg(any(not(feature = "asm"), feature = "asm-aarch64"))]
 pub const K3: u32 = 0xCA62C1D6u32;
 
 pub const H: [u32; STATE_LEN] = [
--- a/sha1/src/lib.rs
+++ b/sha1/src/lib.rs
@@ -28,24 +28,50 @@
 #![no_std]
 #![doc(html_logo_url =
     "https://raw.githubusercontent.com/RustCrypto/meta/master/logo_small.png")]
+
+// Give relevant error messages if the user tries to enable AArch64 asm on unsupported platforms.
+#[cfg(all(feature = "asm-aarch64", target_arch = "aarch64", not(target_os = "linux")))]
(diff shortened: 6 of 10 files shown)


newpavlov:
Thank you for your PRs!

IIUC your assembly code uses instructions from the `crypto` extension. I would be interested to see it implemented using `std` [intrinsics](https://doc.rust-lang.org/core/arch/aarch64/index.html) instead of assembly and compare the resulting performance. It will require usage of a nigthly compiler, so having assembly may be a fine temporary solution, but in future I think we will replace your assembly with a code which does not use the extension instructions.

One thing about which I am not sure is usage of `getauxval`. It's a Linux-specific function, so at the very least you should gate the code appropriately. Ideally we should use AArch64 analogue of `is_x86_feature_detected!`, but since we target no_std and ARM, it's not so easy... We will have to wait for rust-lang/rfcs#2725 to do it properly.

One alternative could be to gate your asm on `target_feature = "crypto"`. Users will have to explicitly enable this feature using compiler flags (or by using `target-cpu=native`) to get the HW acceleration. It would simplify code, but will make it harder to discover this option and will remove the runtime detection, which is not ideal...

linkmauve:
I originally wanted to use intrinsics, but I’d like the [xmpp-rs project](https://gitlab.com/xmpp-rs/xmpp-rs) to work on stable too (and be fast at that).  I’m fine with doing that work again using intrinsics once a path towards stabilisation of AArch64 intrinsics has been decided.  Also note that intrinsics aren’t inherently safer than assembly.  What do you mean by “code which does not use the extension instructions”?  Something which is approximately as slow as the Rust version?

`is_aarch64_feature_detected!()` is also unstable (see https://github.com/rust-lang/rust/issues/27731), so I guess I will expose it only on Linux for now, with a TODO to replace it with the macro once the `stdsimd` feature leaves nightly.

Yeah no, I’d rather have this enabled the same way as on x86, to simplify its usage.  We can migrate to the `target_feature` once both of the above issues have been stabilised.

sha2/src/aarch64.rs
```diff
@@ -0,0 +1,11 @@
+// From sys/auxv.h
+const AT_HWCAP: u64 = 16;
+const HWCAP_SHA2: u64 = 64;
+extern {
+    fn getauxval(type_: u64) -> u64;
+}
+
+pub fn sha1_supported() -> bool {
```
newpavlov: Rename it to `sha2_supported`?

sha2/src/aarch64.rs
```diff
@@ -0,0 +1,11 @@
+// From sys/auxv.h
+const AT_HWCAP: u64 = 16;
+const HWCAP_SHA2: u64 = 64;
+extern {
+    fn getauxval(type_: u64) -> u64;
```
newpavlov: Are you sure we don't need any linker options to use `getauxval`? It may work only because `std` links it as well, so for `no_std` app things could break.

sha2/src/aarch64.rs
```diff
@@ -0,0 +1,11 @@
+// From sys/auxv.h
+const AT_HWCAP: u64 = 16;
+const HWCAP_SHA2: u64 = 64;
```
newpavlov, line 3: BTW is it possible that processor supports SHA-2 instructions, but not SHA-1 and/or AES? I thought they are part of the same `crypto` extension.

newpavlov (commented):
Yeah, let's go with your approach after adding the appropriate gates and replace it sometime in the future after stabilization of the AArch64 intrinsics.

>What do you mean by “code which does not use the extension instructions”? Something which is approximately as slow as the Rust version?

Roughly, yes. IIRC even the current not-so-optimal assembly is a bit faster than the pure Rust version, so I think using OpenSSL assembly will result in a performance boost even when the `crypto` extension is not available. See RustCrypto/asm-hashes#5 for more details.

sha2/src/aarch64.rs
linkmauve: `no_std` only means we don’t use the `std` crate, not that we don’t use the system libc, right?  Do you have a Linux system where this isn’t the case?

sha2/src/aarch64.rs
linkmauve: The “crypto extension” is an ARM marketing term for a Cortex-A core bundling these three extensions, and possibly the sha512, sha3, and more in the future.

I’m not aware of any actual CPU exposing e.g. the sha2 flag but not sha1, but given they are different flags I’d rather test them properly.

sha2/src/aarch64.rs
newpavlov: But you still have to link to `libc`. I think the best approach here will be to use [`libc::getauxval`](https://docs.rs/libc/0.2.66/libc/fn.getauxval.html), which handles correct linking itself. Of course `libc` should be an optional dependency, and we probably should gate your assembly not only on the `asm` feature, but also on the new `libc` feature (or introduce an alias for both of them, e.g. `aarch64-crypto`).

sha2/src/aarch64.rs
newpavlov: Since we are using `getauxval` it does not matter, but during migration to intrinsics we will have to use a more coarse grained target features, which only know about the `crypto` extension. But I guess since LLVM does not introduce separate finer-grained features it should be fine.

linkmauve:
I have also added an ARM worker to the CI, it only fails because there has been no sha1-asm and sha2-asm release containing my code yet (so it wouldn’t work anyway even if you merged it).

sha2/src/aarch64.rs
linkmauve: How does it work for e.g. the sha3 flag?

newpavlov: Honestly, I don't quite get how LLVM handles features, e.g. in [here](http://lists.llvm.org/pipermail/llvm-dev/2018-September/126346.html) they write:
>Extensions can now have different meanings based on the base architecture they apply to. For example on AArch64, 'crypto' means different things for v8.{1,2,3}-a than v8.4-a. The former adds 'sha2' and 'aes', the latter adds those and 'sm4' and 'sha3' on top.

I wonder why they use `crypto` instead of finer-grained `sha1`, `sha2`, `sha512`, `sha3` and `aes` features...

sha2/src/aarch64.rs
linkmauve: https://github.com/rust-lang/cargo/issues/1197 is preventing the obvious solution to enable this feature conditionally, I guess I’ll have to fix it first. :/

Also, libc links to the wrong shared object, it should be libc.so.6 on Linux (so `link = "c"`), not libutil.so… It also doesn’t export `AT_HWCAP` nor the various `HWCAP_*` constants, meh. So many yaks to shave… ^^'

newpavlov: I think for now we can use something like this:
```toml
[dependencies]
libc = { version = "0.2", optional = true }
sha1-asm = { version = "0.4", optional = true }

[features]
asm = ["sha1-asm"]
# has effect only for AArch64 Linux system
aarch64-crypto = ["asm", "libc"]
```

Can you open an issue in the `libc` repo regarding those problems?

linkmauve: I made a PR: https://github.com/rust-lang/libc/pull/1638

linkmauve:
Once https://github.com/RustCrypto/asm-hashes/pull/14 is merged the CI should be green again, after which I think this PR is good to merge. :)

newpavlov:
FYI: crate updates are published now!

sha2/src/aarch64.rs
linkmauve: They apparently noticed they made a mistake and now have aliased crypto to sha2 + aes, see https://github.com/llvm-mirror/llvm/commit/663484452ecf2000001548dfd28c5f816e2aa8b7#diff-5ff776706657bdcfc45aa7d7159bfeadR45

We should be able to use each feature separately nowadays.

PR #104, opened 2020-02-27
Allow compile-time detection of crypto extension on AArch64

linkmauve:
The existing way depends on the `libc` crate, only works on Linux, and does a `getauxval()` call each 64 bytes of input.

This commit adds support for doing this detection at compile-time, if the `"crypto"` `target_feature` is enabled we can use the same code-path as amd64 and reuse the `asm` feature. It should also work on non-Linux `target_os`, but this is untested.

A new `compile_error!()` has been added helping users figure out the best way to build this crate.

Benchmarks don’t seem to change much without the `getauxval()` call, they’re all within error range.


tarcieri:
Thanks!

PR #107, opened 2020-03-21
skein

ghost:


tarcieri:
This looks interesting to potentially include, however it vendors its own copy of Threefish, and should probably be changed to use https://github.com/RustCrypto/block-ciphers/tree/master/threefish (i.e. the `threefish` crate) instead.

tarcieri:
Going to go ahead and close this out as the author seems to no-longer have a GitHub account...

PR #108, opened 2020-03-24
Expose sha256_utils behind a feature flag.

dconnolly:
We need to use `compress256` as part of Zcash Sprout key derivation; this change exposes it as part of the `sha256_utils` module when using the `utils` feature.

### DIFF
--- a/sha2/src/lib.rs
+++ b/sha2/src/lib.rs
@@ -98,12 +98,16 @@ extern crate std;
 mod aarch64;
 mod consts;
 mod sha256;
-#[cfg(any(not(feature = "asm"), feature = "asm-aarch64", feature = "utils"))]
-pub mod sha256_utils;
+#[cfg(any(not(feature = "asm"), feature = "asm-aarch64", feature = "compress"))]
+mod sha256_utils;
 mod sha512;
-#[cfg(any(not(feature = "asm"), target_arch = "aarch64"))]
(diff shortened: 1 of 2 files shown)


sha2/src/lib.rs
```diff
@@ -55,40 +55,55 @@
...
-mod sha512_utils;
 #[cfg(feature = "asm-aarch64")]
 mod aarch64;
+mod consts;
 mod sha256;
+#[cfg(any(not(feature = "asm"), feature = "asm-aarch64", feature = "utils"))]
+pub mod sha256_utils;
 mod sha512;
+#[cfg(any(not(feature = "asm"), target_arch = "aarch64"))]
+mod sha512_utils;
```
tarcieri, line 105: It seems like whatever changes are applied to `sha256_utils` should also be applied to `sha512_utils` for consistency.

tarcieri:
In general this PR has a lot of seemingly unrelated formatting changes which, while good on their own, make it a bit hard to review.

tarcieri:
An alternative suggestion for accomplishing your intended goals here:

How about an off-by-default `compress` cargo feature which re-exports `compress256` and `compress512` from the toplevel?

tarcieri:
Awesome, looks great!

tarcieri (approved):

dconnolly:
Thank you very much! 🎉

PR #109, opened 2020-05-06
sha1, sha2: Use libc for AArch64 consts too

linkmauve:
Now that https://github.com/rust-lang/libc/pull/1638 got merged and a new `libc` release got made (0.2.68), we can use them instead of hardcoding their values.

This bumps the `libc` dependency to the earliest version with these consts.

There is no functional change, just one TODO less.


tarcieri:
Thanks!

PR #153, opened 2020-06-09
Use new `*Dirty` traits from the `digest` crate

tarcieri:
Updates all of the hash implementations in this repo to use the new `*Dirty` traits which support blanket impls of the original traits, which either consume the hasher instance or can be reset.

This will also provide a marginal efficiency boost, at least until placement return lands (which, as it were, may be soon).

### DIFF
--- a/gost94/src/gost94.rs
+++ b/gost94/src/gost94.rs
@@ -3,10 +3,9 @@
 use block_buffer::block_padding::ZeroPadding;
 use block_buffer::byteorder::{ByteOrder, LE};
 use block_buffer::BlockBuffer;
-use digest::generic_array::typenum::U32;
-use digest::generic_array::GenericArray;
 use digest::impl_write;
-use digest::{BlockInput, FixedOutput, Reset, Update};
+use digest::{consts::U32, generic_array::GenericArray};
+use digest::{BlockInput, FixedOutputDirty, Reset, Update};
 
 pub(crate) type Block = [u8; 32];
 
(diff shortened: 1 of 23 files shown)


gost94/src/gost94.rs
```diff
@@ -238,31 +237,30 @@ impl Update for Gost94 {
...
 }
 
-impl FixedOutput for Gost94 {
+impl FixedOutputDirty for Gost94 {
     type OutputSize = U32;
 
-    fn finalize_fixed(mut self) -> GenericArray<u8, U32> {
-        {
-            let self_state = &mut self.state;
+    fn finalize_into_dirty(&mut self, out: &mut GenericArray<u8, U32>) {
```
newpavlov, line 243: I think it could be worth to be consistent and use `digest::Output<Self>` across all crates.

newpavlov (approved):

newpavlov:
Dirty traits fit even better than I expected.

gost94/src/gost94.rs
tarcieri: The reason I couldn't here is because it doesn't `impl Default`, and `digest::Output` (presently, at least) uses the `Digest` trait:

https://github.com/RustCrypto/traits/blob/master/digest/src/digest.rs#L96

It could potentially use `FixedOuput` instead.

tarcieri:
Indeed, surprisingly easy refactor for what it is.

I'll go ahead and merge and circle back on changing `digest::Output` to use `FixedOutput`.

gost94/src/gost94.rs
tarcieri: Hrmm, changing it to use `FixedOutput` is not trivial... it means you can't use it with `Digest` because not all `Digest`s necessarily impl `FixedOutput`, and that kind of defeats the point.

I think we should leave it as-is. That said, I think we can fix the version used in the macro.

PR #162, opened 2020-06-10
ripemd160 v0.9.0

tarcieri:
### Changed
- Update to `digest` v0.9 release ([#155])
- Use new `*Dirty` traits from the `digest` crate ([#153])
- Bump `block-buffer` to v0.8 release ([#151])
- Rename `*result*` to `finalize` ([#148])
- Upgrade to Rust 2018 edition ([#129])

[#155]: https://github.com/RustCrypto/hashes/pull/155
[#153]: https://github.com/RustCrypto/hashes/pull/153
[#151]: https://github.com/RustCrypto/hashes/pull/151
[#148]: https://github.com/RustCrypto/hashes/pull/148
[#129]: https://github.com/RustCrypto/hashes/pull/129

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -210,7 +210,7 @@ checksum = "7e0456befd48169b9f13ef0f0ad46d492cf9d2dbb918bcf38e01eed4ce3ec5e4"
 
 [[package]]
 name = "ripemd160"
-version = "0.9.0-pre"
+version = "0.9.0"
 dependencies = [
  "block-buffer",
  "digest",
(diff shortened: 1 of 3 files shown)


tarcieri:
@newpavlov FYI, going to cut some releases of the higher priority hashes (at least for myself) then circle back on a final release of all of the rest of them

newpavlov:
Ok, right now I am working on updating to the new version of `block-buffer`, I don't think new releases should cause any merge conflicts. Though we will have to release patch updates right after v0.9.0 release.

tarcieri:
If you'd like to do that, I can just cut releases of the ones I immediately need, then we can do patch updates of just those rather than releasing all of them twice.

newpavlov:
I don't mind either, so if you would like to publish updates for other hashes right now, feel free to do so.

PR #166, opened 2020-06-11
[WIP] sha1: add x86 intrinsics support

newpavlov:
On Ryzen 7 2700X I get up to 2.2 GB/s. For now CPUID detection is absent. I plan to add it based on the approach used in `getrandom`.

Unfortunately GitHub Actions workers do not have SHA instructions, so we will not be able to properly test this code in CI.

@linkmauve
Can you please check that I haven't broke ARM code?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -48,6 +48,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7bbb73db36c1246e9034e307d0fba23f9a2e251faa47ade70c1bd252220c8311"
 
 [[package]]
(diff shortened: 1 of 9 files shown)


newpavlov:
Closed in favor of #167.

linkmauve:
Hi, why are you removing runtime feature detection?  Very often the code will be built for a generic CPU and will run on one with those extensions (on ARM at least).

newpavlov:
I am not removing it, it's just not implemented in the draft PR yet.

newpavlov:
And ARM detection works without changes (at least it should).

PR #90, opened 2019-11-05
feat(sha2): implement intrinsic based version for sha256

dignifiedquire:
This is not optimal yet when processing multiple blocks, but it does give a considerable speedup over the non intrinsic version already

Measurements were taken on an `AMD Ryzen 7 3700X 8-Core`.

## Baseline

```
test bench1_10    ... bench:          36 ns/iter (+/- 0) = 277 MB/s
test bench2_100   ... bench:         314 ns/iter (+/- 4) = 318 MB/s
test bench3_1000  ... bench:       3,096 ns/iter (+/- 16) = 322 MB/s
test bench4_10000 ... bench:      30,935 ns/iter (+/- 297) = 323 MB/s
```

## With asm-hashes
```
test bench1_10    ... bench:          33 ns/iter (+/- 0) = 303 MB/s
test bench2_100   ... bench:         291 ns/iter (+/- 1) = 343 MB/s
test bench3_1000  ... bench:       2,863 ns/iter (+/- 14) = 349 MB/s
test bench4_10000 ... bench:      28,597 ns/iter (+/- 191) = 349 MB/s
```

## With Intrinsics

```
test bench1_10    ... bench:           8 ns/iter (+/- 0) = 1250 MB/s
test bench2_100   ... bench:          54 ns/iter (+/- 0) = 1851 MB/s
test bench3_1000  ... bench:         511 ns/iter (+/- 8) = 1956 MB/s
test bench4_10000 ... bench:       5,079 ns/iter (+/- 29) = 1968 MB/s
```

PR #167, opened 2020-06-11
Add x86 intrinsics support for sha1 and sha2

newpavlov:
Written using #90 and the software implementation as a reference.

This PR also simplifies handling of ARM backend and deprecates `asm-aarch64` feature.


dignifiedquire:
@newpavlov I agree, your code looks much nicer :) 

Some thoughts for my specific use cases

(1) Bringing back dynamic detection would be great, often times the code will be built once and then run on different cpus in my case. 
(2) ~~now that `input_blocks` is a thing again it will be more efficient to process multiple blocks at once, like I am doing here: https://github.com/filecoin-project/rust-fil-proofs/blob/master/sha2raw/src/sha256_intrinsics.rs#L12~~ Looks like you already did this
(3) currently all implementations do have some overhead and additional padding in my measurements, which is why I ended up implementing this "raw" construction https://github.com/filecoin-project/rust-fil-proofs/blob/master/sha2raw/src/sha256.rs#L25 which avoids going through the `block_buffer` and can only process appropriately sized and pre padded chunks. It would be great to find a way such that passing in slices of matching data could have the same speed.

Fleshgrinder:
I second the dynamic detection request. This is important for software that is installed on many different systems.

newpavlov:
>Bringing back dynamic detection would be great

The code is written with dynamic dispatch in mind, I just need to write CPUID detection using `getrandom` code as a reference (probably I will move it into a separate crate).

>currently all implementations do have some overhead and additional padding in my measurements

If you don't need padding, then I think the easiest way will be to use the compression functions directly, which can be accessed after enabling the `compress` feature. I guess we also could add imports for initialization constants.

newpavlov:
Runtime detection is back! If you interested, it's powered by the [`cpuid-bool`](https://github.com/RustCrypto/utils/blob/master/cpuid-bool/src/lib.rs) crate.

@linkmauve
Can you verify that crates are not broken on ARM?

tarcieri (approved):

dignifiedquire (approved):
very nice, thank you!

sha1/src/compress/x86.rs
```diff
@@ -0,0 +1,108 @@
...
+    state[1] = _mm_extract_epi32(state_abcd, 2) as u32;
+    state[2] = _mm_extract_epi32(state_abcd, 1) as u32;
+    state[3] = _mm_extract_epi32(state_abcd, 0) as u32;
+    state[4] = _mm_extract_epi32(state_e, 3) as u32;
+}
+
+pub fn compress(state: &mut [u32; 5], blocks: &[[u8; 64]]) {
+    // TODO: Replace with https://github.com/rust-lang/rfcs/pull/2725
+    // after stabilization
+    if cpuid_bool::cpuid_bool!("sha", "sse2", "ssse3", "sse4.1") {
```
jethrogb, line 101: This code breaks platforms that don't support CPUID (e.g. SGX). You should be using https://doc.rust-lang.org/std/macro.is_x86_feature_detected.html, which is the standard way all platforms support feature detection.

newpavlov: Note that `is_x86_feature_detected` is not available in `libcore` and the linked PR proposes changes to change it. See #183 for discussion as how to fix SGX.

PR #168, opened 2020-06-11
Bump opaque-debug to v0.3.0

newpavlov:
Also instead of importing `impl_write!` code now uses `digest::impl_write!` directly.


blake2/src/lib.rs
```diff
@@ -87,9 +87,6 @@
 #![cfg_attr(feature = "simd", feature(platform_intrinsics, repr_simd))]
 #![cfg_attr(feature = "simd_asm", feature(asm))]
 
-#[macro_use]
-extern crate opaque_debug;
```
tarcieri, line 91: 🎉

tarcieri (approved):

PR #172, opened 2020-06-12
Release `gost94` and `streebog`

tarcieri:
Releases the following versions:

- `gost94` v0.8.0
- `streebog` v0.9.0

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -79,7 +79,7 @@ dependencies = [
 
 [[package]]
 name = "gost94"
-version = "0.9.0-pre"
+version = "0.8.0"
 dependencies = [
  "block-buffer",
  "digest",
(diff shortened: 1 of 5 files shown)


tarcieri:
@newpavlov these are both ready to release, right?

newpavlov:
Yes, I believe so. But let's bump `gost94` directly to v0.9.0, so version will be synchronized with `digest.`

newpavlov (approved):

tarcieri:
@newpavlov looks like I don't have access to `gost94`

newpavlov:
I've published `gost94 v0.9.0`and added the team to the list of owners.

tarcieri:
Great! Well all hashes have at least been bumped to `digest` v0.9, but I think there are a few to bump still for the `block-buffer` and `opaque-debug` updates

PR #193, opened 2020-09-28
Fix panic on arithmetic overflow

divergentdave:
I found a crash in the GOST hash function through fuzzing. The test added here was panicking with `attempt to add with overflow`, because there was one place a carry bit was being added with `+` instead of `overflowing_add`.


tarcieri:
Looks good to me

newpavlov (approved):
Thank you! I have modified your PR a bit to use const-time arithmetic. Ideally it should use word-sized integers instead of `u64`s, but the current version should be good enough.

PR #203, opened 2020-11-03
Add enforce-soft feature to sha1 and sha2

### LINKED ISSUE #202: [Feature Request] Disable SIMD with a feature
Hi!

We are developing a new research OS. Even though our OS runs on x86, it currently doesn't support SIMD instructions. It would be nice if there's a feature that allows us to disable the usage of SIMD instructions and use the software implementation instead.

https://github.com/RustCrypto/hashes/blob/bb71874240a8e9f381aa84116168795cedd33357/sha2/src/sha256.rs#L153-L157

newpavlov:
Closes #202

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -223,7 +223,7 @@ dependencies = [
 
 [[package]]
 name = "sha-1"
-version = "0.9.1"
+version = "0.9.2"
 dependencies = [
  "block-buffer",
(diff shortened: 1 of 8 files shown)


tarcieri (approved):
I'm not a huge fan of features that subtract functionality, but this does at least address the immediate issue.

PR #207, opened 2020-11-23
Fix using sha2_asm on aarch64.

lxiange:
The code doesn't compile before.
Can use `cargo test --features asm` to verify (on an aarch64 machine).
Tested on AWS Graviton2 server.

Btw, the current GitHub Action testing scripts failed to cover this case. Because GitHub Actions are hosted on x86 machines.
If want to test on different architectures, maybe we need to setup QEMU to emulate.

### DIFF
--- a/sha2/src/sha256/aarch64.rs
+++ b/sha2/src/sha256/aarch64.rs
@@ -12,9 +12,9 @@ pub fn compress(state: &mut [u32; 8], blocks: &[[u8; 64]]) {
     if sha2_supported() {
         // TODO: replace after sha2-asm rework
         for block in blocks {
-            sha2_asm::compress256(&mut self.h, block);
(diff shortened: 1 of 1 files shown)


tarcieri:
Note: we have pretty extensive `cross`-based testing on the `asm-hashes` repo, including `aarch64`:

https://github.com/RustCrypto/asm-hashes/blob/master/.github/workflows/sha2.yml

lxiange:
> Note: we have pretty extensive `cross`-based testing on the `asm-hashes` repo, including `aarch64`:
> 
> https://github.com/RustCrypto/asm-hashes/blob/master/.github/workflows/sha2.yml

Got it, that's cool!

tarcieri:
Thanks! We should definitely add `cross`-based testing to anything that uses `asm-hashes`

PR #208, opened 2020-11-25
sha2: test builds on 32-bit Linux and macOS/x86-64

tarcieri:
Per #207 it seems we aren't doing enough testing, particularly for the `asm` feature on various platforms.

This commit adds much more extensive testing for the `sha2` crate in particular, since it's one of the most important.


tarcieri:
The Windows and cross-based tests are failing for whatever reason.

That probably warrants some further investigation, however for now we can at least expand the coverage a bit.

PR #217, opened 2021-01-18
Update crates to digest v0.10

### LINKED ISSUE #67: Add `Blake2b-256`, `Blake2b-128`, etc. as wrappers over `VarBlake2b`.

### LINKED ISSUE #317: Add a more advanced version of the "one million a" test
Since blocks are identical  "one million a" may not catch some bugs, e.g. as in #314. One option is to test a long sequence of bytes incrementing from 0 to 100, i.e. `[0, 1, 2, ..., 99, 100, 0, 1, 2, ..., 99, 100, 0, ...]`. Alternatively we could use a simple non-cryptographic PRNG seeded with a fixed seed.

### LINKED ISSUE #330: Update crypto-mac to 0.11 in blake2
Trying to cleanup a dependency tree where we have 3 versions of crypto-mac, one of them from blake2. Would be nice if you could update it.

I tried doing it myself but can't figure out why tests are broken. Here's the diff:

```diff
diff --git a/blake2/Cargo.toml b/blake2/Cargo.toml
index 6d25e15..925a325 100644
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -13,12 +13,12 @@ categories = ["cryptography", "no-std"]

 [dependencies]
 digest = "0.9"
-crypto-mac = "0.8"
+crypto-mac = "0.11.1"
 opaque-debug = "0.3"

 [dev-dependencies]
 digest = { version = "0.9", features = ["dev"] }
-crypto-mac = { version = "0.8", features = ["dev"] }
+crypto-mac = { version = "0.11.1", features = ["dev"] }
 hex-literal = "0.2"

 [features]
diff --git a/blake2/src/blake2.rs b/blake2/src/blake2.rs
index 3255a30..e7ea513 100644
--- a/blake2/src/blake2.rs
+++ b/blake2/src/blake2.rs
@@ -375,7 +375,7 @@ macro_rules! blake2_impl {
                 Self { state }
             }

-            fn new_varkey(key: &[u8]) -> Result<Self, InvalidKeyLength> {
+            fn new_from_slice(key: &[u8]) -> Result<Self, InvalidKeyLength> {
                 if key.len() > $bytes::to_usize() {
                     Err(InvalidKeyLength)
                 } else {
```

newpavlov:
Depends on https://github.com/RustCrypto/traits/pull/819.

Notable changes:
- `ripemd160`, `ripemd256`, and `ripemd320` merged into `ripemd`.
- Complete rewrite of the `groestl` crate using (public domain) code from the NIST submission as a reference. Now the crate has an acceptable performance (~100 MB/s vs ~1 MB/s). It can be improved even further by utilizing AES-NI and SIMD in general, but I don't plan to work on it myself.
- Use LE table for Whirlpool, thus improving performance a bit on LE targets.
- Added Tiger2 variant.
- `fsb` improvements.
- `sha2`, `streebog`, and `shabal` are implemented in terms of the `VariableOutputCore` trait. For example, `Sha512VarCore` implements it, but supports initialization only for 4 sizes and its variants are defined as:
```rust
pub type Sha512_224 = CoreWrapper<CtVariableCoreWrapper<Sha512VarCore, U28>>;
pub type Sha512_256 = CoreWrapper<CtVariableCoreWrapper<Sha512VarCore, U32>>;
pub type Sha384 = CoreWrapper<CtVariableCoreWrapper<Sha512VarCore, U48>>;
pub type Sha512 = CoreWrapper<CtVariableCoreWrapper<Sha512VarCore, U64>>;
```

Closes #67
Closes #317
Closes #330


tarcieri:
FYI, I just merged Tiger in #229, so you'll probably want to rebase

Cargo.toml
```diff
@@ -20,3 +19,6 @@ members = [
 
 [profile.dev]
 opt-level = 2
+
+[patch.crates-io]
+block-buffer = { git = "https://github.com/RustCrypto/utils", branch = "digest_pad" }
```
tarcieri: Note: for whatever reason `cross` seems to have trouble with these directives, so you'll either need to disable the cross-based tests or publish a new release.

There might be other alternatives, like using a `git = "..."` directive in a particular crate's `[dependencies]` section, but I haven't thoroughly investigated.

README.md
```diff
@@ -21,8 +21,7 @@ BLAKE2, SHA-2 or SHA-3.
...
-| `ripemd320` | [RIPEMD-320] | [![crates.io](https://img.shields.io/crates/v/ripemd320.svg)](https://crates.io/crates/ripemd320) |  [![Documentation](https://docs.rs/ripemd320/badge.svg)](https://docs.rs/ripemd320) | [![build](https://github.com/rustcrypto/hashes/workflows/ripemd320/badge.svg?branch=master&event=push)](https://github.com/RustCrypto/hashes/actions?query=workflow:ripemd320+branch:master) | :green_heart:* |
+| `ripemd` | [RIPEMD] | [![crates.io](https://img.shields.io/crates/v/ripemd.svg)](https://crates.io/crates/ripemd) |  [![Documentation](https://docs.rs/ripemd/badge.svg)](https://docs.rs/ripemd) | [![build](https://github.com/rustcrypto/hashes/workflows/ripemd/badge.svg?branch=master&event=push)](https://github.com/RustCrypto/hashes/actions?query=workflow:ripemd+branch:master) | :green_heart: |
```
tarcieri: Nice

Cargo.toml
newpavlov: It's not a problem, I plan to publish `pre` version either way. I use `patch` as a way to verify that introduced API changes fit as intended.

atenjin:
Our bitcoin lib need to be compiled in Wasm, 0.9.0 can't work now... I try 0.10 is ok, hoping this pr could be merged soon...
And thanks for your work all in all.

newpavlov:
0.9 versions should work fine on WASM. If it's a problem with our crates, can you please open an issue with a description and reproduction steps?

newpavlov:
BTW I wonder if we should synchronize all hasher crate versions to v0.10. It would make it much easier to see at a single glance which trait version a crate implements, but it would mean we will have a discontinuity in minor versions for some crates, e.g. `sm3` will jump from v0.3 to v0.10.

PR #224, opened 2021-01-30
Use the SHA-NI extension backend with enabled asm feature

### LINKED ISSUE #220: Use sha2-asm for feature compress on aarch64?
Hi,

We met real performance issue when we tested the application `lotus` (dependent on `sha2` with feature `compress`) on `aarch64` platform.
The root cause is, for `aarch64` only with feature `compress`, the `sha2::compress256` does not use the `aarch64` accelerators, e.g. `sha256su0`, it degradates to `soft_compress`.

We tried three ways to solve this.

A. Add feature `compress, asm` in applications for `aarch64` when using `sha2::compress256`.
To keep `x86` platforms using the intrinsics other than `asm`, modifications to `Cargo.toml` of applications would be:
```
+[target."cfg(target_arch = \"aarch64\")".dependencies]
+sha2 = { version = "0.9.1", features = [ "compress", "asm" ] }
+[target."cfg(target_arch = \"x86_64\")".dependencies]
+sha2 = { version = "0.9.1", features = [ "compress" ] }
```
However, this will enable both "compress" and "asm" no matter on `aarch64` or `x86`.
This is a known BUG of `cargo`, see [1][2], the solution is still unstable.

B. Implement the instrinsic like we did for `x86`. We need `core::arch::aarch64::*` to do this.
It is still unstable for `aarch64`, we investigated the assembly generated for `vsha256su0q_u32`,
there are extra two memory loads before the actual `sha256su0`.
So as of now, the intrinsic can not deliver the best performance as the `asm` did.

C. Use `sha2-asm` for `aarch64` with feature `compress`, like:
```
diff --git a/sha2/Cargo.toml b/sha2/Cargo.toml
index 98bd626..10f9ed1 100644
--- a/sha2/Cargo.toml
+++ b/sha2/Cargo.toml
@@ -35,6 +35,6 @@ hex-literal = "0.2"
 default = ["std"]
 std = ["digest/std"]
 asm = ["sha2-asm", "libc"]
-compress = [] # Expose compress function
+compress = ["sha2-asm", "libc"] # Expose compress function
 force-soft = [] # Force software implementation
 asm-aarch64 = ["asm"] # DEPRECATED: use `asm` instead
diff --git a/sha2/src/sha256.rs b/sha2/src/sha256.rs
index 3eef745..56c5c7a 100644
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -142,7 +142,7 @@ cfg_if::cfg_if! {
     if #[cfg(feature = "force-soft")] {
         mod soft;
         use soft::compress;
-    } else if #[cfg(all(feature = "asm", target_arch = "aarch64", target_os = "linux"))] {
+    } else if #[cfg(all(any(feature = "compress", feature = "asm"), target_arch = "aarch64", target_os = "linux"))] {
         mod soft;
         mod aarch64;
         use aarch64::compress;
```

We want to fix the performance issue via `C`, it would be OK for both `aarch64` and `x86`.
With it, the package is easy to import by applications and delivers same performance with feature `compress` and `asm`.

Suggestions?
Regards.

[1] https://github.com/rust-lang/cargo/issues/2524
[2] https://github.com/rust-lang/cargo/issues/7914

newpavlov:
The SHA extension backend is significantly faster then our ASM code, so with this change on x86 targets we first will check if CPU has the extension and only then we will fallback either to ASM or software implementation.

Closes #220.

cc @dgbo

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -246,7 +246,7 @@ dependencies = [
 
 [[package]]
 name = "sha2"
-version = "0.9.2"
+version = "0.9.3"
 dependencies = [
  "block-buffer",
  "cfg-if",
--- a/sha2/CHANGELOG.md
+++ b/sha2/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## 0.9.3 (2021-01-30)
 ### Changed
-- Use the SHA extension backend with enabled `asm` feature. ([#223])
+- Use the SHA extension backend with enabled `asm` feature. ([#224])
 
-[#223]: https://github.com/RustCrypto/hashes/pull/223
+[#224]: https://github.com/RustCrypto/hashes/pull/224
 
 ## 0.9.2 (2020-11-04)
 ### Added
--- a/sha2/CHANGELOG.md
+++ b/sha2/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## 0.9.3 (2021-01-30)
+### Changed
+- Use the SHA extension backend with enabled `asm` feature. ([#223])
+
+[#223]: https://github.com/RustCrypto/hashes/pull/223
+
 ## 0.9.2 (2020-11-04)
 ### Added
 - `force-soft` feature to enforce use of software implementation. ([#203])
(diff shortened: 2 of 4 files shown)


not-an-aardvark:
Thanks for making this change. Would it make sense to do the same thing with SHA1 in addition to SHA2? I'm in a similar situation where I'd like to use the SHA1 intrinsic if available, otherwise fall back to the asm implementation. As-is it seems like the `asm` feature disables the intrinsics.

not-an-aardvark:
I went ahead and created https://github.com/RustCrypto/hashes/pull/225 to apply the same change to SHA1.

dgbo:
Thank you! This does solve our problem.

PR #225, opened 2021-01-31
Use SHA1 intrinsics when `asm` feature is enabled

not-an-aardvark:
This applies the change from https://github.com/RustCrypto/hashes/pull/224 to the SHA1 crate (it was just added to the SHA2 crate before).


newpavlov:
Thank you! I will release the update right away.

PR #50, opened 2018-02-15
Add Tiger hash

jmcomets:
Note that this depends on the addition of a `len_padding_with` method to
`BlockBuffer*` structs (RustCrypto/utils#7).

PR #229, opened 2021-02-04
add tiger hash from https://gitlab.com/rawler/tiger

myers:
Add the tiger hash.

The bulk of this is from [Ulrik Mikaelsson](https://gitlab.com/rawler/tiger) but also work from [Igor Raits](https://gitlab.com/ignatenkobrain/tiger/-/tree/upgrade) and [myself](https://gitlab.com/myerscarpenter/tiger/-/tree/digest-0.9) in upgrading to newer versions of the Digest crate.

@tarcieri suggested adding this crate to this repo so that it can be updated as needed.


tiger/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "tiger"
+version = "0.1.2"
```
tarcieri: I'd suggest setting this to:

```suggestion
version = "0.0.0"
```

We can bump it on release.

tiger/src/lib.rs
```diff
@@ -0,0 +1,1255 @@
+//! The [Tiger][1] hash function, implemented through [Digest][2]
+//!
+//! [1]: http://www.cs.technion.ac.il/~biham/Reports/Tiger/tiger/tiger.html
+//! [2]: https://crates.io/crates/digest
+
```
tarcieri, line 5: It looks like you should be able to add `no_std` and our other attribute boilerplate without issue:

```suggestion
#![no_std]
```

tarcieri:
Looks like you need to `rustfmt`

tiger/Cargo.toml
```diff
@@ -0,0 +1,25 @@
...
+description = "Tiger hash function"
+authors = ["Ulrik Mikaelsson <ulrik.mikaelsson@gmail.com>", "RustCrypto Developers"]
+license = "Apache-2.0"
```
tarcieri: All of our other crates are `MIT OR Apache-2.0`. Is that acceptable?

myers: that's a good question.  The original work is released under Apache 2.0.  From reading https://law.stackexchange.com/questions/6081/can-i-bundle-mit-licensed-components-in-a-apache-2-0-licensed-project it seems like adding this under the MIT license is easy?  

Should we ask Ulirk?

myers:
based on the failure on thumbv7em-none-eabi the byte-order crate doesn't work under nostd.  Looks like what it does you can use byte-tools instead.  I'll work on that.

tiger/Cargo.toml
tarcieri: It would definitely be good to get the original author's blessing to republish it as MIT+Apache2.0

.github/workflows/tiger.yml
```diff
@@ -0,0 +1,60 @@
...
+          - wasm32-unknown-unknown
+    steps:
+      - uses: actions/checkout@v1
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          target: ${{ matrix.target }}
+          override: true
+      - run: cargo build --no-default-features --release --target ${{ matrix.target }}
```
tarcieri, line 41: These won't pass until you add the `#![no_std]` attribute

tiger/Cargo.toml
myers: I have emailed him.  I started to make it look like it was under both licenses, but removed that.

myers:
@tarcieri I think I have addressed everything you mentioned above.   

If the license question is an issue, I could also go and resurrect #50

tarcieri:
The license issue is pretty important as we try to ensure our crates all have consistent licensing and a clear IPR story.

We can wait to hear from the author or you can revive #50.

Perhaps you can benchmark the two and see if there's a compelling reason other than licensing to go with one over the other?

myers:
> From: Ulrik Mikaelsson 
> Date: Fri, 5 Feb 2021 07:35:53 +0100
> Subject: Re: Relicense rust tiger under MIT?
> 
> Fun to hear you're using it! I'm ok with licensing it under MIT, but i also
> got notice the the PR is now closed in favor of other PR, so not sure it
> matters?

I will make the changes to have it listed as under MIT OR Apache 2.0

myers:
benchmark of this PR's tiger crate:

```
$ cd tiger
$ cargo bench
[...]
test bench1_10    ... bench:          20 ns/iter (+/- 3) = 500 MB/s
test bench2_100   ... bench:         158 ns/iter (+/- 17) = 632 MB/s
test bench3_1000  ... bench:       1,678 ns/iter (+/- 240) = 595 MB/s
test bench4_10000 ... bench:      15,735 ns/iter (+/- 2,818) = 635 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 11.33s
```

tiger/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2016 felipeamp
```
tarcieri: Hmm, that seems like it might not be right?

myers:
I have spent 20 minutes to try to get #50 to compile and pass tests so that I can benchmark, but it's failing tests. 

I propose this PR is ready for merging.

tarcieri:
Looks good, thank you!

I'll cut a 0.1 after merging.

tarcieri:
It's released: https://crates.io/crates/tiger

PR #238, opened 2021-02-16
sha1: expose compression function

### LINKED ISSUE #237: Expose sha1::compress publicly?
I'm trying to accomplish the following task on [this project](https://github.com/not-an-aardvark/lucky-commit):

1. Using the `sha-1` crate, run sha1 on a few blocks of data, once, without finalizing.
2. Save the sha1 state vector from step (1)
3. Run a different implementation of sha1 (e.g. within a GPU) on a small number of remaining blocks, using the state vector from (2) as an initial state.

This is a bit difficult to do with the `sha-1` library. Since a `Sha1` struct doesn't expose its internal state (the `h` field), it's not possible to copy the state from step 2 and use it elsewhere. Additionally, since the `Sha1` struct is the only entrypoint to the crate, I also can't use `sha1::compress` to keep track of state while buffering the data manually. My current workaround is to depend on the `sha1-asm` crate, which [does expose the `compress` API I need](https://docs.rs/sha1-asm/0.5.0/sha1_asm/fn.compress.html). But this isn't ideal, because I lose support for the `sha-1` crate's implementation that uses SHA1 intrinsics, as well as its software implementation of `compress` for architectures that don't support the assembly version.

Admittedly, this is a very unusual use case, so it would be understandable if the answer is to just fork the repo. But since `sha1::compress` is a fairly well-established primitive and is already implemented, I thought I would ask it you'd be open to making it accessible from outside the crate.

newpavlov:
Closes #237

### DIFF
--- a/.github/workflows/sha1.yml
+++ b/.github/workflows/sha1.yml
@@ -54,3 +54,4 @@ jobs:
       - run: cargo test --no-default-features
       - run: cargo test
       - run: cargo test --features asm
+      - run: cargo test --all-features
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -223,7 +223,7 @@ dependencies = [
 
 [[package]]
 name = "sha-1"
-version = "0.9.3"
+version = "0.9.4"
 dependencies = [
  "block-buffer",
  "cfg-if",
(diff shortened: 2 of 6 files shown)


tarcieri (approved):
It might be good to add some `doc_cfg` to this crate and also `sha2`

PR #239, opened 2021-02-23
blake2.rs: remove unnecessary `unsafe`

abonander:
`.copy_from_slice()` is mildly annoying to use since it panics if the slices aren't the same length but using `ptr::copy_nonoverlapping()` is also a code smell.

This should perform the same as the previous version (including the `assert!()`) as the optimizer should be able to see that the internal assert in `copy_from_slice()` is always satisfied.


abonander:
The Clippy error looks to be unrelated to this change.

newpavlov:
Note that we plan completely rewrite `blake2`, see #228. Plus in #217 this method is already removed. So I think we can close this PR in favor of those two.

PR #240, opened 2021-02-24
Update cpuid-bool dependency to 0.2

jack-signal:


jack-signal:
Clippy failure seems unrelated

tarcieri:
Fixed clippy in #241

tarcieri:
Thanks!

PR #249, opened 2021-03-30
sm3: add OSCCA SM3 secure hash

uudiin:
Add OSCCA SM3 secure hash (OSCCA GM/T 0004-2012 SM3) generic
hash transformation.


sm3/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "sm3"
```
tarcieri, line 2: It looks like there is already a published `sm3` crate:

https://crates.io/crates/sm3

The repo/source code is here:

https://github.com/andelf/rust-sca/tree/master/sm3

Perhaps we could ask the author about upstreaming it.

uudiin: Putting these crates on RustCrypto would be a better choice, I will try to contact the original author. Thanks.

andelf: I've invited @newpavlov as an owner of `sm3` on crates.io.

Feel free to overwrite it with RustCrypto crates. 😁

Also, I've invited @uudiin to `rust-sca` repo. You can update any READMEs and links.

tarcieri: Wow, awesome, thank you!

tarcieri:
@newpavlov can you add the `RustCrypto/hashes` team as an owner of the `sm3` crate?

newpavlov:
@tarcieri 
Done!

PR #256, opened 2021-04-24
FSB hash function

iquerejeta:
As an exercise, I've implemented the FSB hash function taking as a reference the paper, so I might have missed out on some efficiency tricks in the reference implementation. 

If you'd be happy to have this code in the repo, I'm happy to maintain it and make the necessary changes to take it to the repo quality standards!


.gitignore
```diff
@@ -1,3 +1,5 @@
 target/
 */target/
 */*/target/
+.idea
+.DS_Store
```
tarcieri: Speaking as a fellow IntelliJ user...

I'd suggest adding the following to your personal `~/.gitconfig` instead:

```
[core]
  excludesfile = /home/myuser/.gitignore
```

...and then putting these kinds of directives in there.

That avoids every other project having to add `.gitignore` directives for every editor.

fsb/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "fsb_rust"
```
tarcieri: RustCrypto owns the `fsb` name. You can use that.

@newpavlov can you add `rustcrypto:hashes` as an owner?

fsb/Cargo.toml
```diff
@@ -0,0 +1,16 @@
...
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[dependencies]
+whirlpool = "0.9"
+digest = "0.9.0"
+block-buffer = { version = "0.9", features = ["block-padding"] }
+hex-literal = "0.3.1"
+
+[dev-dependencies]
+rand = "0.7.3"
```
tarcieri: Would be good to bump this to v0.8.

tarcieri:
Looks like a promising start!

@iquerejeta can you rebase? You'll probably want to `git checkout --theirs Cargo.lock` and have it recomputed from there.

fsb/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+#![allow(non_snake_case)]
+#[allow(dead_code)]
+#[macro_use]
+mod macros;
+mod pi;
```
tarcieri, line 59: GitHub won't even let me leave comments on that module, but... wow 😮

tarcieri:
@iquerejeta it looks like you need to bump clippy MSRV in `.github/workflows/workspace.yml` L19

Also can copy/modify a README.md from one of the other crates? (e.g. `blake2`) Thanks!

fsb/src/lib.rs
iquerejeta: Not my work! Copied that file from the reference implementation

iquerejeta:
I've rebased, hopefully correctly 😄  
- [x] Missing the bump on clippy

qq, should I squash all commits to have a cleaner history?

tarcieri:
We use GitHub's "Squash and merge" feature so you don't need to worry about squashing it yourself

iquerejeta:
Bumped version. I got a security audit failing in the repo, but can't really tell what the problems are. Also I got a bunch of PRs from the bot, requesting to bump dependencies on other crates (e.g. sha1 or sha2). Should I go ahead and merge all these PRs? 
Thanks!

tarcieri:
Tests appear to be passing now.

.github/workflows/workspace.yml
```diff
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.41.0 # MSRV
+          toolchain: 1.47.0 # MSRV
```
newpavlov: I think we can remove the "MSRV" comment from here, since it's not about MSRV, but about pinning clippy version.

Cargo.lock
```diff
@@ -81,6 +81,10 @@ dependencies = [
  "generic-array",
 ]
 
+[[package]]
+name = "fsb_rust"
+version = "0.1.0"
```
newpavlov, line 86: Since you've already changed the crate name, re-generate the lock file, so it will use `fsb` instead of `fsb_rust`? I have shared ownership of the `fsb` crate with the hashes group, so we can publish it immediately after merge.

fsb/Cargo.toml
```diff
@@ -0,0 +1,20 @@
...
+authors = ["RustCrypto Developers"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+edition = "2018"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["crypto", "fsb", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+
+[dependencies]
+whirlpool = "0.9"
```
newpavlov: I suggest using `whirlpool = { version = "0.9", path = "../whirlpool", default-features = false }` here.

fsb/README.md
```diff
@@ -0,0 +1,58 @@
+# RustCrypto: FSB
+
+<!---
+[![crate][crate-image]][crate-link]
+[![Docs][docs-image]][docs-link]
+[![Build Status][build-image]][build-link]
+-->
```
newpavlov: Don't forget to uncomment these lines?

fsb/src/lib.rs
```diff
@@ -0,0 +1,21 @@
+#![allow(non_snake_case)]
+#[allow(dead_code)]
+#[macro_use]
```
newpavlov, line 57: Is it truly necessary to allow dead code? It also would be nice to add crate-level docs, `no_std` and other attributes (see other crates for reference).

fsb/tests/lib.rs
```diff
@@ -0,0 +1,146 @@
...
+    let msg_3 = b"tiriri tralala potompompom";
+
+    let mut hasher = Fsb160::new();
+    hasher.update(msg_1);
+    let result = hasher.finalize();
+
+    assert_eq!(
+        result[..],
+        hex!("6e8ce7998e4c46a4ca7c5e8f6498a5778140d14b")[..]
+    );
```
newpavlov, line 15: You can simply use the following code here:
```rust
assert_eq!(
    Fsb160::digest(msg_1)[..],
    hex!("6e8ce7998e4c46a4ca7c5e8f6498a5778140d14b")[..],
);
```

newpavlov (commented):
I haven't checked the implementation itself, so I have only several surface-level comments for now.

fsb/src/lib.rs
iquerejeta: added `no_std`, and I depend on `alloc` now for dynamically sized arrays. We use them [here](https://github.com/iquerejeta/hashes/blob/master/fsb/src/macros.rs#L37).

tarcieri:
@iquerejeta thank you!

iquerejeta:
Pleasure! I'm happy to change the tests using the `use digest::new_test;` macro, but when I tried it out I couldn't figure out how to generate the data (e.g. `hashes/sha2/tests/data/sha256.blb`), any tips?

PR #267, opened 2021-05-06
sha2: use `cpufeatures` to detect `sha2` on `aarch64`

tarcieri:
Uses the newly added `aarch64` support in the `cpufeatures` crate for `sha2` CPU feature detection on Linux and macOS/M1.

### DIFF
--- a/sha2/Cargo.toml
+++ b/sha2/Cargo.toml
@@ -21,11 +21,12 @@ opaque-debug = "0.3"
 cfg-if = "1.0"
 sha2-asm = { version = "0.6.1", optional = true }
 
-[target.'cfg(any(target_arch = "x86", target_arch = "x86_64"))'.dependencies]
-cpufeatures = "0.1"
-
+[target.aarch64-apple-darwin.dependencies]
+cpufeatures = { version = "0.1", git = "https://github.com/rustcrypto/utils.git" }
 [target.'cfg(all(target_arch = "aarch64", target_os = "linux"))'.dependencies]
-libc = { version = "0.2.93", optional = true }
+cpufeatures = { version = "0.1", git = "https://github.com/rustcrypto/utils.git" }
+[target.'cfg(any(target_arch = "x86", target_arch = "x86_64"))'.dependencies]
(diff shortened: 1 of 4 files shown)


sha2/Cargo.toml
```diff
@@ -21,11 +21,12 @@ opaque-debug = "0.3"
 cfg-if = "1.0"
 sha2-asm = { version = "0.6.1", optional = true }
 
-[target.'cfg(any(target_arch = "x86", target_arch = "x86_64"))'.dependencies]
-cpufeatures = "0.1"
-
+[target.aarch64-apple-darwin.dependencies]
```
dignifiedquire, line 24: seems to be missing linux based aarch64

sha2/Cargo.toml
```diff
@@ -21,11 +21,12 @@ opaque-debug = "0.3"
...
+cpufeatures = { version = "0.1", git = "https://github.com/rustcrypto/utils.git" }
 [target.'cfg(all(target_arch = "aarch64", target_os = "linux"))'.dependencies]
```
tarcieri, line 26: Linux-based `aarch64` is right here 😉

sha2/Cargo.toml
tarcieri: 👇

tarcieri:
Confirm that the tests pass and this successfully detects `sha2` support on an M1 Mac Mini:

```
test bench1_10    ... bench:           6 ns/iter (+/- 0) = 1666 MB/s
test bench2_100   ... bench:          47 ns/iter (+/- 2) = 2127 MB/s
test bench3_1000  ... bench:         463 ns/iter (+/- 2) = 2159 MB/s
test bench4_10000 ... bench:       4,639 ns/iter (+/- 30) = 2155 MB/s
```

tarcieri:
Unfortunately it doesn't work on Linux because the `cpufeatures` macro is expanding code containing references to `libc`, which is a bug.

I will go ahead and extract that code into a small helper function ala the MacOS implementation.

Edit: fix here https://github.com/RustCrypto/utils/pull/398

PR #276, opened 2021-05-15
sha1+sha2: make aarch64 CPU feature detection OS-independent

tarcieri:
`cpufeatures` v0.1.4 can build on any `aarch64` target, not just Linux/macOS ones:

https://github.com/RustCrypto/utils/pull/40

This PR bumps the minimum `cpufeatures` requirement to v0.1.4 and removes the OS-specifisms on `aarch64` targets.

On platforms other than Linux/macOS, CPU features are always detected as `false` unless the corresponding `target_feature`s have been enabled via `RUSTFLAGS`. But in the future, we can add new platform support to `cpufeatures` without changing all of its downstream consumers.


dignifiedquire:
nice

PR #277, opened 2021-05-15
Make documentation more precise

ineiti:
According to https://en.wikipedia.org/wiki/SHA-2#Comparison_of_SHA_functions, it's 8 x 32 bits for sha256 and 8 x 64 bits for sha512. Reading just `the 32-bit 'Sha256'` is confusing to me.

But YMMV...


tarcieri:
I agree. Thanks!

PR #282, opened 2021-06-03
Use archive.org link to Whirlpool description

nabijaczleweli:
The original one 404s, this is the one used by Wikipedia


tarcieri:
Thanks!

PR #285, opened 2021-07-04
sha1: pass block slices to sha1_asm + remove cfg toggle for constants

tlercher:
This doesn't really depend on RustCrypto/asm-hashes#38 but it is required to make it work for me on a M1 Mac.

As the title says i changed the (internal) API so that block slices instead of independent blocks are passes to `sha1_asm`.

I also removed the `cfg` toggles for `aarch64`, because they are required for the software-fallback and `asm-aarch64` is marked deprecated anyway.

Benchmarks on my M1 MacBookPro:

## Soft
```
test bench1_10    ... bench:          14 ns/iter (+/- 0) = 714 MB/s
test bench2_100   ... bench:         120 ns/iter (+/- 1) = 833 MB/s
test bench3_1000  ... bench:       1,151 ns/iter (+/- 4) = 868 MB/s
test bench4_10000 ... bench:      11,465 ns/iter (+/- 130) = 872 MB/s
```

## ASM
```
test bench1_10    ... bench:           7 ns/iter (+/- 0) = 1428 MB/s
test bench2_100   ... bench:          45 ns/iter (+/- 1) = 2222 MB/s
test bench3_1000  ... bench:         459 ns/iter (+/- 9) = 2178 MB/s
test bench4_10000 ... bench:       4,579 ns/iter (+/- 84) = 2183 MB/s
```

### DIFF
--- a/sha1/src/compress.rs
+++ b/sha1/src/compress.rs
@@ -15,9 +15,7 @@ cfg_if::cfg_if! {
         #[cfg(feature = "asm")]
         mod soft {
             pub(crate) fn compress(state: &mut [u32; 5], blocks: &[[u8; 64]]) {
-                for block in blocks {
-                    sha1_asm::compress(state, block);
-                }
+                sha1_asm::compress(state, blocks);
             }
         }
         mod x86;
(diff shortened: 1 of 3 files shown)


tarcieri:
Oh whoops, didn't see this in time, but the same changes were made in #289. Thanks anyway!

PR #298, opened 2021-07-19
ripemd256: add CI config

tarcieri:


Cargo.lock
```diff
@@ -138,9 +138,9 @@ dependencies = [
 
 [[package]]
 name = "hex-literal"
-version = "0.3.1"
+version = "0.3.3"
```
newpavlov: Change to `hex-literal` v0.2?

PR #301, opened 2021-08-04
blake2: make `simd_opt` work on nightly again

Centril:
By using `inline_const`'s `const { $e }` expressions for `simd_shuffleN`s.


tarcieri:
It seems this only compiles on nightly and breaks parsing on other releases

Centril:
Think I fixed the issue.

tarcieri:
Thanks!

Issue #311, opened 2021-09-06
sha2: intrinsics based AVX2 backend


newpavlov:
RustCrypto/asm-hashes#41 adds an assembly based AVX2 backend. It may be worth looking into porting it to intrinsics. Assuming it will have a similar performance to the assembly, we may remove the latter in favor of the former.

newpavlov:
Probably solved in #312.

PR #312, opened 2021-09-06
Add x86 intrinsics support for sha2-512

Rexagon:
This version is a port of this implementation (`src/sha512_compress_x86_64_avx2.c`) [https://github.com/aws-samples/sha2-with-c-intrinsic](https://github.com/aws-samples/sha2-with-c-intrinsic) which is based on this document [https://eprint.iacr.org/2012/067.pdf](https://eprint.iacr.org/2012/067.pdf). It uses `avx` impl for single odd block and then `avx2` for the rest.

### Benchmarks:

CPU: Ryzen 9 5950X

`cargo +nightly bench -p sha2 --features force-soft`

```
     Running unittests (target/release/deps/sha512-0451eb22388b5224)

running 4 tests
test bench1_10    ... bench:          22 ns/iter (+/- 0) = 454 MB/s
test bench2_100   ... bench:         184 ns/iter (+/- 8) = 543 MB/s
test bench3_1000  ... bench:       1,869 ns/iter (+/- 5) = 535 MB/s
test bench4_10000 ... bench:      18,226 ns/iter (+/- 54) = 548 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 5.02s
```

`cargo +nightly bench -p sha2`

```
     Running unittests (target/release/deps/sha512-fc38283933ec7290)

running 4 tests
test bench1_10    ... bench:          18 ns/iter (+/- 0) = 555 MB/s
test bench2_100   ... bench:         148 ns/iter (+/- 1) = 675 MB/s
test bench3_1000  ... bench:       1,317 ns/iter (+/- 11) = 759 MB/s
test bench4_10000 ... bench:      12,836 ns/iter (+/- 156) = 779 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 3.06s
```


Rexagon:
Hopefully related to #311

tarcieri (approved):
Looks good to me.

@newpavlov WDYT?

sha2/src/sha512/x86.rs
```diff
@@ -0,0 +1,354 @@
...
+    if avx2_cpuid::get() {
+        unsafe {
+            sha512_compress_x86_64_avx2(state, blocks);
+        }
+    } else {
+        super::soft::compress(state, blocks);
+    }
+}
+
+#[target_feature(enable = "avx,avx2,sse2,sse3")]
```
newpavlov: AVX2 [implies](https://doc.rust-lang.org/reference/attributes/codegen.html#x86-or-x86_64) AVX. And this chain goes further: AVX2 -> AVX -> SSE4.2 -> SSE4.1 -> SSSE3 -> SSE3 -> SSE2. In other words, you only need to enable `avx2`.

sha2/src/consts.rs
```diff
@@ -163,6 +163,56 @@ pub const K64X2: [[u64; 2]; 40] = [
...
 
+macro_rules! dup_array {
+    ([$([$a:expr, $b:expr]),*,]) => {[
+        $($b, $a, $b, $a),*,
+    ]}
+}
+
+/// Constants necessary for SHA-512 family of digests.
+pub const K64X4: [u64; 160] = dup_array!([
+    [K64[1], K64[0]],
```
newpavlov: I wonder if we can remove this duplication without negatively impacting performance. i think we can do necessary shuffling in one-two instructions. Yes, it's more than zero, but it would mean less memory pressure.

Also if we are to keep this table, then we could enforce it's alignment to 32 bytes, so we will be able to use `_mm256_load_si256` (aligned load) instead of `_mm256_loadu_si256` (unaligned load). Ideally this table would contain `__m256i`, but unfortunately `_mm256_set_epi64x` is not `const`ified, so we would have to use `transmute_copy` for it.

newpavlov (commented):
I haven't looked deep into implementation itself and would like to try some minor tweaks, but I think we can merge it after the `target_feature(enable = "..")` nitpick will be fixed.

sha2/src/consts.rs
Rexagon: Checked with this wrapper for alignment:
```rust
#[repr(align(32))]
pub struct Align32;

pub struct AlignedArray<A, T> {
    _alignment: [A; 0],
    value: T,
}

impl<A, T> core::ops::Deref for AlignedArray<A, T> {
    type Target = T;

    #[inline]
    fn deref(&self) -> &T {
        &self.value
    }
}

pub const K64X4: AlignedArray<Align32, [u64; 160]> = AlignedArray { .... };
```

The difference is insignificant so removed this array completely

newpavlov:
Thank you! I will soon release it as v0.9.7

PR #313, opened 2021-09-08
Release sha2 v0.9.7

newpavlov:


jplatte:
@newpavlov @tarcieri This release broke Sha384, please yank!

I created a minimal reproduction as a private repo and invited you both to see it. It might only affect some inputs (there is a working test for sha384 in this repo afterall). This breakage was brought to my attention in SQLx'es Discord (not sure whether [this link](https://discord.com/channels/665528275556106240/665528275556106243/885300653679861810) works) because SQLx hashes migration scripts to make sure exactly that migration was applied.

tarcieri:
I've gone ahead and yanked this out of caution.

I will investigate and see if I can reproduce the problem. If I can, I'll file a @RustSec advisory.

newpavlov:
Oh, sorry about that!

`force-soft` works without issues and you don't use the `asm` feature, correct?

cc @Rexagon

tarcieri:
FWIW I wasn't able to reproduce this problem

newpavlov:
@tarcieri 
Just to be sure: are you running it on an x86-64 CPU with AVX2?

tarcieri:
Yes, Intel Core i9

newpavlov:
I was able to reproduce this bug. Looks like there is a problem in the AVX2 part. If we are processing blocks one by one, then the AVX code works correctly.

tarcieri:
Aah, it's possible the input message I used to test wasn't long enough

PR #314, opened 2021-09-09
sha2: Fix bug in the AVX2 backend

newpavlov:
The backend was introduced in #312. The bug was reported in https://github.com/RustCrypto/hashes/pull/313#issuecomment-915666732.

AVX part of the backend works correctly, the issue arises only in the AVX2 part (i.e. when we process two blocks at once). Interestingly enough this bug was not caught by the "1 million a" test for SHA-512, so this bug requires two different blocks. Our tests for SHA-384 and SHA-512 are probably too short for exposing this bug (we need a message of at least 256 bytes for that). 

For now I only added a reproduction test. I will try to fix this bug a bit later.

cc @jplatte @Rexagon

### DIFF
--- a/sha2/CHANGELOG.md
+++ b/sha2/CHANGELOG.md
@@ -5,7 +5,7 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## 0.9.7 (2021-09-08)
+## 0.9.7 (2021-09-08) [YANKED]
 ### Added
 - x86 intrinsics support for SHA-512 ([#312])
 
--- a/sha2/src/sha512/x86.rs
+++ b/sha2/src/sha512/x86.rs
@@ -106,8 +106,8 @@ unsafe fn load_data_avx2(
 
     macro_rules! unrolled_iterations {
         ($($i:literal),*) => {$(
-            x[$i] = _mm256_insertf128_si256(x[$i], _mm_loadu_si128(data.add($i) as *const _), 1);
-            x[$i] = _mm256_insertf128_si256(x[$i], _mm_loadu_si128(data.add($i + 1) as *const _), 0);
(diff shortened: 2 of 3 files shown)


Rexagon:
Found the imposter in `load_data_avx2`:
```rust
x[$i] = _mm256_insertf128_si256(x[$i], _mm_loadu_si128(data.add($i) as *const _), 1);
x[$i] = _mm256_insertf128_si256(x[$i], _mm_loadu_si128(data.add($i + 1) as *const _), 0);
```

should be:
```rust
x[$i] = _mm256_insertf128_si256(x[$i], _mm_loadu_si128(data.add(8 + $i) as *const _), 1);
x[$i] = _mm256_insertf128_si256(x[$i], _mm_loadu_si128(data.add($i) as *const _), 0);
```

How could tests be extended to capture a little bit more cases?

newpavlov:
Thank you! Your patch indeed fixes the bug.

>How could tests be extended to capture a little bit more cases?

To be extra sure that the AVX2 and software backends are identical to each other I've wrote the following code:
```rust
let mut rng = rand::thread_rng();
let mut state1 = [0u64; 8];
let mut state2;
let mut blocks = [[0u8; 128]; 256];
for i in 0.. {
    if i % 1000 == 0 {
        println!("{:?}", i);
    }
    rng.fill(&mut state1);
    state2 = state1.clone();
    let n = rng.gen::<u8>() as usize;
    for block in blocks[..n].iter_mut() {
        rng.fill(block);
    }
    soft::compress(&mut state1, &blocks[..n]);
    avx2::compress(&mut state2, &blocks[..n]);
    assert_eq!(state1, state2);
}
```
With the updated version it run for 10M iterations without any panics.

I think we should replace the "one million a" tests with something more advanced, e.g. by testing long sequence of numbers iterating from 0 to 100. But I think I will do it in a separate PR.

tarcieri:
> I think we should replace the "one million a" tests with something more advanced, e.g. by testing long sequence of numbers iterating from 0 to 100. But I think I will do it in a separate PR.

This might be a good use case for some equivalence testing, perhaps using proptest as the driver

PR #320, opened 2021-10-06
Remove some `unsafe`

noncombatant:
As far as I can tell, building locally, these functions don't need to be marked `unsafe`.


tarcieri:
Thanks!

PR #325, opened 2021-11-12
Add cSHAKE128 and cSHAKE256 implementations

elichai:
A few notes about the implementation:
1. I left the code for changing the function name (N) commented out, I can either remove it, or leave it with an empty function name (although that might degrade the performance a little bit).
2. I added another `Sha3State` so `CSHAKE` can implement `Reset` without allocating the customization string. but because of that I couldn't use the state macro.
3. I added a `has_customization` boolean that if it's not it makes it the same as SHAKE.
4. I had to copy some logic from `digest::dev` because nothing there could support a hash function that doesn't implement Default (We could prob extend `xof_test` to accept a constructor, but we'll need a different `new_test!` macro that will support Blob3 and know to call `new` on the hasher with the customization)

The test vectors are from https://github.com/damaki/libkeccak/tree/master/tests/kat/testvectors/cSHAKE together with the current SHAKE tests for an empty customization.


elichai:
I'm starting to think it might be better to remove the `Reset` implementation so that not everyone will need to pay for that extra State just for the few people that actually use it.

newpavlov:
Sorry for the late reply! Can you please rebase and update the code to `digest` v0.10?

elichai:
Superceeded by https://github.com/RustCrypto/hashes/pull/355

PR #336, opened 2021-12-14
README: Fix chained example wrt digest v0.10

aewag:
With digest v0.10 the `Digest::chain()` method was renamed to `Digest::chain_update()`.


newpavlov:
Thank you!

PR #338, opened 2021-12-15
Sha3 refactor unsafe

aewag:
I was looking into the sha3 implementation and would like to propose this PR to avoid the use of unsafe.

While the implementation differentiated between little & big endian, I ended for the little endian path with more or less similar code than it was existing in the big endian path already. Thus, I assume that there is no differentiation needed. Unfortunately, I could only test it on my machine, which is LE.

Performance-wise it does not really change anything, see below first without unsafe as it is now and second as it was before:

850bcc2 sha3: Add forbid(unsafe_code) crate attribute
```
test sha3_224_10    ... bench:          38 ns/iter (+/- 3) = 263 MB/s
test sha3_224_100   ... bench:         370 ns/iter (+/- 19) = 270 MB/s
test sha3_224_1000  ... bench:       3,706 ns/iter (+/- 582) = 269 MB/s
test sha3_224_10000 ... bench:      36,332 ns/iter (+/- 2,908) = 275 MB/s
test sha3_256_10    ... bench:          39 ns/iter (+/- 3) = 256 MB/s
test sha3_256_1000  ... bench:       3,894 ns/iter (+/- 125) = 256 MB/s
test sha3_256_10000 ... bench:      38,367 ns/iter (+/- 4,029) = 260 MB/s
test sha3_265_100   ... bench:         398 ns/iter (+/- 76) = 251 MB/s
test sha3_384_10    ... bench:          52 ns/iter (+/- 31) = 192 MB/s
test sha3_384_100   ... bench:         515 ns/iter (+/- 77) = 194 MB/s
test sha3_384_1000  ... bench:       5,151 ns/iter (+/- 1,037) = 194 MB/s
test sha3_384_10000 ... bench:      50,180 ns/iter (+/- 2,378) = 199 MB/s
test sha3_512_10    ... bench:          74 ns/iter (+/- 4) = 135 MB/s
test sha3_512_100   ... bench:         730 ns/iter (+/- 69) = 136 MB/s
test sha3_512_1000  ... bench:       7,212 ns/iter (+/- 269) = 138 MB/s
test sha3_512_10000 ... bench:      73,020 ns/iter (+/- 8,302) = 136 MB/s
test shake128_10    ... bench:          33 ns/iter (+/- 2) = 303 MB/s
test shake128_100   ... bench:         319 ns/iter (+/- 10) = 313 MB/s
test shake128_1000  ... bench:       3,234 ns/iter (+/- 187) = 309 MB/s
[...]

PR #339, opened 2021-12-16
sha3: Add cross tests

aewag:
Well, that escalated quickly... Sorry for any noise! 

This PR adds cross-tests to `sha3` crate. It does so by using the composite action named `cross-tests`, which can be found in `.github/actions/cross-tests/action.yml`

Inputs to the `cross-tests` actions are: Rust Version, Package, Target & Features

I adapted the `sha1` & `sha2` crate to use the same composite action to reduce duplicated code (other crates seemed not to use any cross-tests).

Further I changed the place where `MSRV` is defined. It now can be set in the `env` section in the beginning of the `.yml` file at a single place.
Disadvantage: It needs one job run before anything else to distribute it. But that could also be used in the future to create the Strategy Matrix at a single place and distribute it to the jobs. 

Issue came up in https://github.com/RustCrypto/hashes/pull/338

### DIFF
--- a/.github/actions/cross-tests/action.yml
+++ b/.github/actions/cross-tests/action.yml
@@ -1,0 +1,35 @@
+name: "cross-tests"
+
+inputs:
+  rust:
+    required: true
+  package:
+    required: true
+  target:
+    required: true
+  features:
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ inputs.rust }}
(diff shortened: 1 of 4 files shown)


tarcieri (approved):
Interesting. TIL composite actions.

Looks good to me, but I'll let @newpavlov review as well.

tarcieri:
Going to merge this. It seems good for `cross`-checking crates, and with the composite action maybe we can wire up additional checks for other crates (e.g. `blake2`)

aewag:
Thanks. :)
I don't know, if other RustCrypto crates could benefit from `cross-tests` or are already running some, but if so it could make sense to keep generic composite actions, such as `cross`, stored in a single repo.

PR #345, opened 2022-01-06
sha2: fix AVX backend

### LINKED ISSUE #344: Bug in Sha512 AVX(2) implementation
I found a bug in Sha512 AVX2 implementation and I'm not sure what's the nicest fix.
to trigger the bug I ran the following code:
```rust
use sha2::{Digest, Sha512};
fn main() {
    let mut h = Sha512::new();
    h.update(&[0u8; 300]);
    println!("{:?}", h.finalize());
}
```

With these flags: `RUSTFLAGS='-Clink-arg=-fuse-ld=lld -Clinker-plugin-lto' cargo r`
Output:
`Segmentation fault (core dumped)`

The source of the bug is here: https://github.com/RustCrypto/hashes/blob/c478cbb/sha2/src/sha512/x86.rs#L117
`_mm_store_si128` is used with a pointer to a u64 array, which is aligned to 8 and not to 16 as the instruction requires.
(there are a lot of places that assume that `MsgSchedule` and `RoundStates` are aligned to 16 bytes while this is in fact false)

We need to make them aligned correctly, there are a few ways I see how this can be done:
1. Make 2 structs with `repr(align(16))` and implement Index/IndexMut with all the `Range*` variants.
2. Make 2 structs with `repr(align(16))` and transmute their mutable reference to a mutable reference to u64's.
3. Make a `Align16` type and replace these with arrays of that type, and add a function to receive the lower and higher part of that type.

(we can cut boilerplate of 1 in half via const generics)

Would appreciate thoughts before I implement anything as the first one has a lot of boilerplate code in it.

(P.S. You might notice that  `&mut ms[2 * $i] as *mut u64 as *mut _` is also a bug because it takes a mutable reference to a single u64 and casts it into a pointer with size 2*64, so reading from that violates Stack Borrows, but that's easier to fix via `ms[2 * $i.. 2 * $i + 2].as_mut_ptr() as *mut u64 as *mut _ `)

newpavlov:
Closes #344

### DIFF
--- a/sha2/src/sha512/x86.rs
+++ b/sha2/src/sha512/x86.rs
@@ -34,8 +34,8 @@ unsafe fn sha512_compress_x86_64_avx2(state: &mut [u64; 8], blocks: &[[u8; 128]]
         start_block += 1;
     }
 
-    let mut ms: MsgSchedule = Default::default();
-    let mut t2: RoundStates = [0u64; SHA512_ROUNDS_NUM];
+    let mut ms: MsgSchedule = [_mm_setzero_si128(); 8];
+    let mut t2: RoundStates = [_mm_setzero_si128(); 40];
     let mut x = [_mm256_setzero_si256(); 8];
 
     for i in (start_block..blocks.len()).step_by(2) {
@@ -56,7 +56,7 @@ unsafe fn sha512_compress_x86_64_avx2(state: &mut [u64; 8], blocks: &[[u8; 128]]
 
 #[inline(always)]
 unsafe fn sha512_compress_x86_64_avx(state: &mut [u64; 8], block: &[u8; 128]) {
-    let mut ms = Default::default();
+    let mut ms = [_mm_setzero_si128(); 8];
     let mut x = [_mm_setzero_si128(); 8];
 
     // Reduced to single iteration
@@ -82,7 +82,7 @@ unsafe fn load_data_avx(x: &mut [__m128i; 8], ms: &mut MsgSchedule, data: *const
                 _mm_loadu_si128(&K64[2 * $i] as *const u64 as *const _),
             );
 
-            _mm_store_si128(&mut ms[2 * $i] as *mut u64 as *mut _, y);
+            ms[$i] = y;
         )*};
     }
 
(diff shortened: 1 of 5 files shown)


sha2/src/sha512/x86.rs
```diff
@@ -341,9 +336,17 @@ fn_sha512_update_x!(sha512_update_x_avx2, __m256i, {
         XOR = _mm256_xor_si256,
 });
 
+fn cast_ms(ms: &MsgSchedule) -> &[u64; SHA512_BLOCK_WORDS_NUM] {
+    unsafe { &*(ms as *const MsgSchedule as *const _) }
+}
+
+fn cast_rs(rs: &RoundStates) -> &[u64; SHA512_ROUNDS_NUM] {
+    unsafe { &*(rs as *const RoundStates as *const _) }
+}
```
elichai: nit, `#[inline(always)]`?

elichai:
This is much nicer!

elichai:
On my machine(i9-9980HK) with turbo boost off, ran it a few times to make sure I get consistent result

Before:
```
test sha256_10    ... bench:          80 ns/iter (+/- 6) = 125 MB/s
test sha256_100   ... bench:         808 ns/iter (+/- 99) = 123 MB/s
test sha256_1000  ... bench:       7,734 ns/iter (+/- 402) = 129 MB/s
test sha256_10000 ... bench:      76,839 ns/iter (+/- 24,415) = 130 MB/s
test sha512_10    ... bench:          42 ns/iter (+/- 1) = 238 MB/s
test sha512_100   ... bench:         405 ns/iter (+/- 12) = 246 MB/s
test sha512_1000  ... bench:       3,532 ns/iter (+/- 587) = 283 MB/s
test sha512_10000 ... bench:      33,786 ns/iter (+/- 651) = 295 MB/s
```

After:
```
test sha256_10    ... bench:          80 ns/iter (+/- 1) = 125 MB/s
test sha256_100   ... bench:         789 ns/iter (+/- 11) = 126 MB/s
test sha256_1000  ... bench:       7,727 ns/iter (+/- 28) = 129 MB/s
test sha256_10000 ... bench:      76,364 ns/iter (+/- 1,379) = 130 MB/s
test sha512_10    ... bench:          40 ns/iter (+/- 0) = 250 MB/s
test sha512_100   ... bench:         390 ns/iter (+/- 22) = 256 MB/s
test sha512_1000  ... bench:       3,497 ns/iter (+/- 117) = 285 MB/s
test sha512_10000 ... bench:      33,845 ns/iter (+/- 7,469) = 295 MB/s
```

elichai:
Just ran the same code in #344 and it doesn't crash
```
$ RUSTFLAGS='-Clink-arg=-fuse-ld=lld -Clinker-plugin-lto' cargo r
    Finished dev [unoptimized + debuginfo] target(s) in 0.01s
     Running `target/debug/wat`
[114, 100, 141, 235, 57, 20, 22, 206, 85, 130, 13, 65, 63, 33, 253, 46, 205, 24, 97, 124, 98, 149, 88, 88, 5, 158, 217, 104, 21, 56, 14, 150, 12, 43, 176, 5, 247, 183, 146, 200, 42, 155, 182, 2, 46, 191, 224, 193, 58, 249, 124, 43, 226, 194, 67, 166, 58, 22, 38, 122, 42, 128, 55, 218]
```

PR #351, opened 2022-01-18
Reduce amount of unsafe and use new imports

newpavlov:
Depends on https://github.com/RustCrypto/traits/pull/849

### DIFF
--- a/.github/workflows/sha1.yml
+++ b/.github/workflows/sha1.yml
@@ -137,11 +137,10 @@ jobs:
 
   # Cross-compiled tests
   cross:
-    needs: set-msrv
     strategy:
       matrix:
         rust:
-          - ${{needs.set-msrv.outputs.msrv}}
+          - 1.51 # 1.41-1.50 `--features` can't be used inside virtual manifest
           - stable
         target:
           - aarch64-unknown-linux-gnu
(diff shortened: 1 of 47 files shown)


.github/workflows/sha1.yml
```diff
@@ -137,11 +137,10 @@ jobs:
 
   # Cross-compiled tests
   cross:
-    needs: set-msrv
     strategy:
       matrix:
         rust:
-          - ${{needs.set-msrv.outputs.msrv}}
+          - 1.51 # 1.41-1.50 `--features` can't be used inside virtual manifest
```
aewag, line 143: This was the reason why to change the directory and not to use the package option [1].

[1] https://github.com/RustCrypto/hashes/pull/351/commits/1902274667add95b2fe469c2e184f42a1c1aa7cd#diff-7989ad9cb1e81934c64205efcfae23eb8bfc6aa827cc5c0f992a9704687a56e8L31

newpavlov: Addition of this option was intentional. It is needed in cases when we prototype stuff using `[patch.crates-io]` in workspace's Cargo.toml, without it `cross` ignores it.

aewag: Ah, makes sense :+1:. I did not have this in mind.

PR #363, opened 2022-02-16
Fix and test min versions build

newpavlov:


aewag:
Would 'minimal-versions' maybe also a candidate to be within RustCrypto/actions?

newpavlov:
@aewag 
Yes, I plan to do it a bit later.

tarcieri:
@aewag I suggested the same thing here, as it were: https://github.com/RustCrypto/block-ciphers/pull/303#pullrequestreview-884742751

newpavlov:
I couldn't port the minimal versions job to a composite action (see https://github.com/RustCrypto/actions/pull/3), so I think I will leave it for a separate PR when the action will be ready.

@aewag 
It would be great if you could handle it since it looks like you are more proficient with GitHub Actions than me.

sha2/Cargo.toml
```diff
@@ -15,22 +15,22 @@ keywords = ["crypto", "sha2", "hash", "digest"]
...
-hex-literal = "0.2"
+digest = { version = "0.10.3", features = ["dev"] }
+hex-literal = "0.2.2"
 
 [features]
 default = ["std"]
 std = ["digest/std"]
-asm = ["sha2-asm"]
-compress = [] # Expose compress function
+asm = ["sha2-asm"] # WARNING: this feature SHOULD NOT be enabled by library crates
```
nazar-pc, line 32: Can you clarify why was this warning added?

newpavlov: Because if a crate somewhere deep in your dependency tree enables it, it could become virtually impossible to disable it. Considering that the assembly crates may fail to build on some targets (e.g. see https://github.com/RustCrypto/asm-hashes/issues/34), it can be quite unpleasant for users downstream.

In future we may migrate to the configuration flags approach which we now use in some cipher crates, e.g. see the [relevant `aes` docs](https://docs.rs/aes/latest/aes/#configuration-flags).

nazar-pc: Link to the issue is exactly what I was looking for, thanks!

PR #366, opened 2022-02-18
blake2: Fix bug on BE machines

aewag:
This PR addresses https://github.com/RustCrypto/hashes/issues/356. Additionally, it adds a cross job to the blake2 workflow with a LE & BE machine.


newpavlov:
Thank you!

PR #374, opened 2022-04-06
Use correct variable name in Base64 hash example

rsaarelm:


tarcieri:
Thanks. We should probably set up something to CI these examples using rustdoc.

PR #375, opened 2022-04-08
Update workspace feature resolver version to 2

zeegomo:
Version 1 of feature resolver always causes features unification of
build and dev dependencies, which is unwanted in some circumstances.
For example, the `dev` feature of the `digest` crate always depends on
`alloc` which requires to define a global allocator in static libs
even though that is never used by library code.

This might be useful in other repos as well

### DIFF
--- a/blake2/src/simd.rs
+++ b/blake2/src/simd.rs
@@ -15,6 +15,7 @@ pub use self::simdty::{u32x4, u64x4};
 pub trait Vector4<T>: Copy {
     fn gather(src: &[T], i0: usize, i1: usize, i2: usize, i3: usize) -> Self;
 
+    #[allow(clippy::wrong_self_convention)]
     fn from_le(self) -> Self;
     fn to_le(self) -> Self;
 
--- a/.github/workflows/workspace.yml
+++ b/.github/workflows/workspace.yml
@@ -17,7 +17,7 @@ jobs:
       - uses: RustCrypto/actions/cargo-cache@master
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: 1.47.0
+          toolchain: 1.60.0
           components: clippy
           profile: minimal
           override: true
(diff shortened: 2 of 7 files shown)


newpavlov:
~~Isn't it default for 2021 edition crates?~~

UPD: Ah, I forgot we haven't migrated crates in this repository yet.

zeegomo:
Seems like the clippy job is failing because the toolchain used is an old one which does not yet support the new resolver on master. Is this (or migrating to the 2021 edition) something you would be interested in given a MSRV bump is likely needed?

tarcieri:
You can bump clippy to whatever version is needed (and really we should probably keep it close to stable to maximize the lints it performs)

zeegomo:
I've bumped clippy to 1.60 to get the latest lints. This uncovered a bunch of warnings which I have currently fixed at the best of my judgement without deep knowledge of the codebase, maybe those should be better handled in a different PR.
However, the `resolver` entry in `Cargo.toml` is ignored by all MSRV builds, even though that does not cause any failures.

blake2/src/simd.rs
```diff
@@ -15,7 +15,7 @@ pub use self::simdty::{u32x4, u64x4};
 pub trait Vector4<T>: Copy {
     fn gather(src: &[T], i0: usize, i1: usize, i2: usize, i3: usize) -> Self;
 
-    fn from_le(self) -> Self;
+    fn from_le(vec: Self) -> Self;
```
tarcieri: Not sure why this was changed, but it's breaking the build:

https://github.com/RustCrypto/hashes/runs/5969024622?check_suite_focus=true#step:3:352

zeegomo: Clippy complained about  a `from_*` method taking self as a parameter, so I changed it to be an associated function but forgot to update it everywhere

tarcieri: I think you can just disable the warning for now. FWIW we plan on switching to a different implementation eventually: #228

tarcieri:
Thanks!

> This might be useful in other repos as well

We have it enabled in certain repos, but if you see others that don't have it on, feel free to add it.

PR #385, opened 2022-08-10
Implement SerializableState for hashes

rpiasetskyi:
The internal state of different hashes can be used for serialization.


sha2/src/lib.rs
```diff
@@ -72,15 +73,50 @@ pub use sha512::compress512;
...
+pub type Sha256CoreSerializedState =
+    GenericArray<u8, <Sha256Core as HashCoreSerializableState>::SerializedStateSize>;
+
+/// SHA-512/224 core.
+pub type Sha512_224Core = CtVariableCoreWrapper<Sha512VarCore, U28>;
+/// SHA-512/224 core.
+pub type Sha512_224 = CoreWrapper<Sha512_224Core>;
+/// SHA-512/224 core serialized state.
+pub type Sha512_224CoreSerializedState =
+    GenericArray<u8, <Sha512_224Core as HashCoreSerializableState>::SerializedStateSize>;
```
newpavlov: I don't think we need to introduce a whole bunch of type aliases for serialized states.

sha2/tests/mod.rs
```diff
@@ -1,14 +1,61 @@
...
+use digest::{
+    crypto_common::SerializableState,
+    dev::{feed_rand_16mib, fixed_reset_test},
+    new_test,
+};
 use hex_literal::hex;
 use sha2::{Digest, Sha224, Sha256, Sha384, Sha512, Sha512_224, Sha512_256};
 
+macro_rules! test_serializable_state {
+    ($name:ident, $hasher:ty) => {
```
newpavlov: You should add tests which check stability of the serialized state, i.e. that every version of this crate results in the same serialized state for a given hasher state.

sha2/tests/mod.rs
```diff
@@ -1,14 +1,61 @@
...
+        #[test]
+        fn $name() {
+            let mut h = <$hasher>::new();
+
+            // First part of data.
+            feed_rand_16mib(&mut h);
```
newpavlov: 16 MiB is not a great value for testing serialization, we should use an odd number of bytes to check that buffer gets serialized properly. Also it's a bit too big in my opinion, we can simply use small fixed strings with length equal to block size plus several bytes.

sha2/tests/mod.rs
rpiasetskyi: Test macro is moved to `traits`, added `expected_serialized_state`. https://github.com/RustCrypto/traits/pull/1078/commits/793f9f268d220be0420e2c6b285a367ade048f88.

sha2/tests/mod.rs
rpiasetskyi: Test macro is moved to `traits`, input data is changed to `BlockSize + 1`. 
https://github.com/RustCrypto/traits/pull/1078/commits/793f9f268d220be0420e2c6b285a367ade048f88.

rpiasetskyi:
Could you please suggest the best solution for tests if the expected values are very long?

```
digest::hash_serialization_test!(md5_serialization, Md5, 
    hex!("e5ca2295db93d6dd07ab990fcad2219e00000000000000010113000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000"));
```
```
digest::hash_serialization_test!(
    md5_serialization,
    Md5,
    [
        0xe5, 0xca, 0x22, 0x95, 0xdb, 0x93, 0xd6, 0xdd, 0x07, 0xab, 0x99, 0x0f, 0xca, 0xd2, 0x21,
        0x9e, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x01, 0x01, 0x13, 0x00, 0x00, 0x00, 0x00,
        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
        0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00
    ]
);
```

newpavlov:
You can split it into several lines:
```rust
let bytes = hex!(
    "0001020304050607"
    "08090a0b0c0d0e0f"
);
// equivalent to:
let bytes = hex!("000102030405060708090a0b0c0d0e0f");
```
It requires `hex-literal v0.3.3`.

rpiasetskyi:
Is it okay if I bump `hex-literal`? This change is moving MSRV of crates to 1.45.
If yes, what is the preferable width of the line?

newpavlov:
Ah, I forgot about that. In this case simply use new lines:
```rust
let bytes = hex!("
    0001020304050607
    08090a0b0c0d0e0f
");
```
Unfortunately, rustfmt does not like it too much, so you may need to add `#[rustfmt::ignore]` or use an alternative formatting choice.

md5/src/lib.rs
```diff
@@ -124,6 +129,38 @@ impl fmt::Debug for Md5Core {
...
 
+impl SerializableState for Md5Core {
+    type SerializedStateSize = U24;
+
+    fn serialize(&self) -> SerializedState<Self> {
+        let mut serialized_state = GenericArray::<u8, U16>::default();
+
+        for (i, bytes) in serialized_state.as_mut_slice().chunks_mut(4).enumerate() {
+            bytes.copy_from_slice(self.state[i].to_be_bytes().as_slice());
+        }
```
newpavlov, line 151: In Rust it's usually preferred to `zip` iterators in such cases instead of using `enumerate`.
```rust
let mut state = GenericArray::<u8, U16>::default();
for (val, chunk) in self.state.iter().zip(state.chunks_exact_mut(4)) {
    chunk.copy_from_slice(&val.to_le_bytes());
}
// You also can transform arrays into GenericArray and back using `From`/`Into`
state.concat(self.block_len.to_le_bytes().into());
```

Also little-endian would be more efficient on most currently used CPUs, so I think we should prefer it. (BE is used traditionally by cryptographers, but it begins to change slowly)

md2/src/lib.rs
```diff
@@ -146,5 +153,24 @@ impl fmt::Debug for Md2Core {
...
+        GenericArray::<u8, U48>::from(self.x).concat(self.checksum)
+    }
+
+    fn deserialize(
+        serialized_state: &SerializedState<Self>,
+    ) -> Result<Self, DeserializeStateError> {
+        let (serialized_x, serialized_checksum) = Split::<u8, U48>::split(serialized_state);
+
+        Ok(Self {
+            x: serialized_x.as_slice().try_into().unwrap(),
```
newpavlov: `serialized_x.into()` does not work here, correct?

rpiasetskyi: Thanks! `(*serialized_x).into()` works. Also, I have changed it for `block_len` in other algorithms.

gost94/tests/mod.rs
```diff
@@ -11,6 +11,71 @@ new_test!(
...
+    hex!("
+        51aeb30f746350e15ef31472e3914b1b
+        4b9198e0272881ff2401cea8490e5ab2
+        00010000000000000000000000000000
+        00000000000000000000000000000000
+        13131313131313131313131313131313
+        13131313131313131313131313131313
+        01130000000000000000000000000000
+        00000000000000000000000000000000
+        00
```
newpavlov, line 27: Hm... This redundant byte for eager hashes is certainly an eyesore. It could be worth to add `serialize`/`deserialize` methods to the `block-buffer` crate.

tarcieri:
Merged as #574

PR #388, opened 2022-08-20
sha-1: fix compilation on ARM32 with `asm` feature (and other unsupported archs)

davidkna:
`sha1-asm` only builds on `aarch64`, `x86` and `x86_64`


tarcieri (approved):

newpavlov:
Looks good! Could you do the same for `md-5`, `sha2`, and `whirlpool`?

davidkna:
@newpavlov Done!

PR #399, opened 2022-09-02
Fix MSRV issues by re-releasing sha1, sha2, sha3, md5, and whirlpool

newpavlov:
Closes #394
Closes #395
Closes #396
Closes #397
Closes #398

PR #401, opened 2022-09-02
Fix MSRV issues by re-releasing sha1, sha2, sha3, md5, and whirlpool. Take 2.

newpavlov:
#399 has not resolved the issues. `resolver = "2"` leaks during publishing using the current stable toolchain and other older toolchains, so this PR temporarily comments it out. ~~It could be a cargo bug, so I reported it in https://github.com/rust-lang/cargo/issues/11047.~~ The behavior looks intentional.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -144,7 +144,7 @@ checksum = "8371e4e5341c3a96db127eb2465ac681ced4c433e01dd0e938adbef26ba93ba5"
 
 [[package]]
 name = "md-5"
-version = "0.10.3"
+version = "0.10.4"
 dependencies = [
  "digest",
  "hex-literal",
@@ -192,7 +192,7 @@ dependencies = [
 
 [[package]]
 name = "sha1"
-version = "0.10.3"
+version = "0.10.4"
 dependencies = [
  "cfg-if",
  "cpufeatures",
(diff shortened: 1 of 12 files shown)


newpavlov:
@striezel Can you confirm that the issues got resolved?

striezel:
@newpavlov Yes, looks good to me. The new versions of the published crates can be built with rustc 1.48 again. :+1:
Thank you very much.

striezel:
Bad news:
`gost94` v0.10.2 (https://docs.rs/crate/gost94/0.10.2/source/Cargo.toml) and `fsb` v0.1.2 (https://docs.rs/crate/fsb/0.1.2/source/Cargo.toml) also have the `resolver = "2"` line, so they will fail to build with older rustc versions, too.
Should I create issues for those, too?

Good news:
`blake2` v0.10.4 (https://docs.rs/crate/blake2/0.10.4/source/Cargo.toml) is __not__ affected by this issue, i. e. builds just fine. That is the MSRV problem likely got introduced _after_ 18th February 2022 (release of blake2 v0.10.4), and earlier releases are not affected. So all other RustCrypto/hashes crates should be fine.

newpavlov:
I will do re-releases for them right away, so I think issues are not necessary. Thank you for checking the other crates!

PR #406, opened 2022-09-10
ripemd: implement RIPEMD-128

lumag:
For the sake of completeness, implement the RIPEMD-128 digest algorithm.


newpavlov (approved):
Thank you!

PR #410, opened 2022-09-20
update actions used in GitHub Actions workflows

striezel:
Updates the actions [`actions/checkout`](https://github.com/actions/checkout) and [`actions/cache`](https://github.com/actions/cache) to v3, their latest major release.

This should basically be backwards compatible, so I expect no breakage.

### DIFF
--- a/.github/actions/cross-tests/action.yml
+++ b/.github/actions/cross-tests/action.yml
@@ -13,7 +13,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: RustCrypto/actions/cargo-cache@master
       - uses: actions-rs/toolchain@v1
         with:
--- a/.github/workflows/blake2.yml
+++ b/.github/workflows/blake2.yml
@@ -29,7 +29,7 @@ jobs:
           - thumbv7em-none-eabi
           - wasm32-unknown-unknown
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - uses: RustCrypto/actions/cargo-cache@master
       - uses: actions-rs/toolchain@v1
         with:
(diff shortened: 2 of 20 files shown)


striezel:
The failed check for macOS (https://github.com/RustCrypto/hashes/actions/runs/3091706476/jobs/5002197600) seems to be caused by network errors:

```
Updating crates.io index
warning: spurious network error (2 tries remaining): SecureTransport error: connection closed via error; class=Net (12)
warning: spurious network error (1 tries remaining): unexpected http status code: 429; class=Http (34)
error: failed to get `digest` as a dependency of package `blake2 v0.10.4 (/Users/runner/work/hashes/hashes/blake2)`

Caused by:
  failed to fetch `[https://github.com/rust-lang/crates.io-index`](https://github.com/rust-lang/crates.io-index%60)

Caused by:
  network failure seems to have happened
  if a proxy or similar is necessary `net.git-fetch-with-cli` may help here
  https://doc.rust-lang.org/cargo/reference/config.html#netgit-fetch-with-cli

Caused by:
  unexpected http status code: 429; class=Http (34)
Error: Process completed with exit code 101.
```

Maybe re-running that check later may help.

newpavlov:
Yeah, we get it sometimes. I've already restarted the failed jobs.

tarcieri:
We might consider adding dependabot configs to all repos to ensure other actions are up-to-date, similar to this one:

https://github.com/RustCrypto/formats/pull/710/files

I think I opened an issue about this somewhere (traits?) but can't find it ATM

striezel:
Usually actions do not get major updates that often, maybe once a year or so. And even if they get a major version update, it is rarely critical with regards to security. So adding dependabot configurations for actions may be a bit of an overkill.

On the other hand, dependabot configurations for Rust package dependencies are a completely different thing. Rust crates get updated much more frequently, and having them up to date is much more important. And you already have those here.

tarcieri:
We have ~30 repos to update every time actions we use are updated. Almost all of them are currently out-of-date.

Some automation might be helpful there.

PR #413, opened 2022-09-21
md5: Add OID support

lumag:


newpavlov:
Can you please bump `md5` version to v0.10.5 with release date set to 2022-09-22? I will do release right after merge. Do not forget to update root Cargo.lock and changelog.

Issue #408, opened 2022-09-16
ripemd: which OIDs to use?


newpavlov:
Unfortunately, ripemd has several OIDs. For example, for Ripemd160 has ISO (1.0.10118.3.0.49) and Teletrust (1.3.36.3.2.1). Go's std uses the former (see https://github.com/golang/go/issues/35495), while OpenSSL the latter.

I am inclined to use the latter in our codebase, because it looks more widespread and the ISO range contains OIDs for Streebog (see http://oid-info.com/get/1.0.10118.3.0), even though AFAIK in practice people usually use the [TC26 OIDs](https://tc26.ru/about/protsedury-i-reglamenty/identifikatory-obektov-oid-tekhnicheskogo-komiteta-po-standartizatsii-kriptograficheskaya-zashchita-1.html) for it.

lumag:
In #415 I've used the Teletrust one. It is clearly specified so in the RFC 4880 text. The OpenSSL would also use it for the generated signatures. GnuTLS also uses the Teletrust OID.

newpavlov:
@tarcieri 
If you don't have any objections, I plan to merge #415.

PR #415, opened 2022-09-22
ripemd: Add OID support

lumag:

### DIFF
--- a/.github/workflows/ripemd.yml
+++ b/.github/workflows/ripemd.yml
@@ -49,7 +49,9 @@ jobs:
     strategy:
       matrix:
         rust:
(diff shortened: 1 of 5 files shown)


newpavlov:
We need to have an explicit decision on the issue described in #408 before implementing OID support for RIPEMD hashes.

lumag:
@newpavlov let's have this PR open and move the discussion to the mentioned issue. Marking this PR as a draft to prevent merging.

newpavlov:
Regardless of our choice, I think it's worth to add a note about existence of multiple OIDs with reference to the issue.

lumag:
@newpavlov done

PR #417, opened 2022-10-03
gost94: add OID support

lumag:
This is an RFC for the 'oid' feature. I do not have example files with the DSTU usage, but hopefully @IvashchenkoSerhii can comment whether `1.2.804.2.1.1.1.1.2.1` is used as a separate digest parameter OID or if it is an OID of the GOST 34.311-95 with the DSTU sBox.

Additional notice. I have added `AssociatedOid` implementation for the individual params, howeve I do not see a way to get such OID given only the upper-level Gost94 type (`CoreWrapper<Gost945Core<Gost94Params>>`). Suggestions would be appreciated.

### DIFF
--- a/gost94/src/params.rs
+++ b/gost94/src/params.rs
@@ -1,3 +1,6 @@
+#[cfg(feature = "oid")]
+use digest::const_oid::{AssociatedOid, ObjectIdentifier};
+
 pub(crate) type Block = [u8; 32];
 pub(crate) type SBox = [[u8; 16]; 8];
 
@@ -29,6 +32,12 @@ impl Gost94Params for CryptoProParam {
     ];
     const H0: Block = [0; 32];
     const NAME: &'static str = "Gost94CryptoPro";
+}
+
+#[cfg(feature = "oid")]
+impl AssociatedOid for CryptoProParam {
+    /// Per RFC 4357
+    const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.643.2.2.30.1");
 }
 
 /// S-Box defined in GOST R 34.12-2015.
@@ -67,6 +76,12 @@ impl Gost94Params for TestParam {
     ];
     const H0: Block = [0; 32];
     const NAME: &'static str = "Gost94Test";
+}
+
+#[cfg(feature = "oid")]
+impl AssociatedOid for TestParam {
+    /// Per RFC 4357
+    const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.643.2.2.30.0");
 }
 
 /// S-Box defined in GOST 34.311-95 & GOST 28147:2009.
--- a/.github/workflows/gost94.yml
+++ b/.github/workflows/gost94.yml
@@ -49,7 +49,7 @@ jobs:
     strategy:
       matrix:
         rust:
-          - 1.41.0 # MSRV
+          - 1.57.0 # MSRV
           - stable
     steps:
       - uses: actions/checkout@v3
@@ -63,3 +63,21 @@ jobs:
       - run: cargo test --no-default-features
       - run: cargo test
       - run: cargo test --all-features
+
+  # TODO: merge with test on MSRV bump to 1.57 or higher
+  test-msrv-41:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.41.0 # MSRV
+    steps:
+      - uses: actions/checkout@v3
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+      - run: cargo test
+      - run: cargo test --no-default-features
--- a/gost94/Cargo.toml
+++ b/gost94/Cargo.toml
@@ -15,9 +15,10 @@ categories = ["cryptography", "no-std"]
 digest = "0.10.3"
 
 [dev-dependencies]
-digest = { version = "0.10.3", features = ["dev"] }
+digest = { version = "0.10.4", features = ["dev"] }
 hex-literal = "0.2.2"
 
 [features]
 default = ["std"]
 std = ["digest/std"]
+oid = ["digest/oid"] # Enable OID support. WARNING: Bumps MSRV to 1.57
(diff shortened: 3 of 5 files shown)


gost94/src/gost94_core.rs
```diff
@@ -196,6 +198,15 @@ impl<P: Gost94Params> Gost94Core<P> {
...
 
+#[cfg(feature = "oid")]
+#[cfg_attr(docsrs, doc(cfg(feature = "oid")))]
+impl<P> AssociatedOid for Gost94Core<P>
+where
+    P: Gost94Params,
+{
+    const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.643.2.2.9");
+}
+
```
newpavlov: I would expect something like:
```rust
impl<P> AssociatedOid for Gost94Core<P>
where
    P: Gost94Params + AssociatedOid,
{
    const OID: ObjectIdentifier = P::OID;
}
```

Or maybe even:
```rust
impl AssociatedOid for Gost94Core<TestParam> {
    const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.643.2.2.30.0");
}

impl AssociatedOid for Gost94Core<CryptoProParam> {
    const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.643.2.2.30.1");
}
```

The latter approach would not be extensible for parameters defined in third-party crates, but I am not sure we need such capability.

newpavlov (commented):
I wonder how the [retracted TC26 OID](https://tc26.ru/about/protsedury-i-reglamenty/identifikatory-obektov-oid-tekhnicheskogo-komiteta-po-standartizatsii-kriptograficheskaya-zashchita-1.html) `1.2.643.7.1.1.2.1` relates to the ones used in this PR. IIUC the OIDs are introduced by the CryptoPro company and it's not clear how "official" they are.

gost94/src/gost94_core.rs
lumag: The problem is that the GOST 34.311-95 is referenced using the `1.2.643.2.2.9` OID rather than individual params OIDs. See, for example, the RFC4490 and its examples. Or likewise the examples from MR 26.2.002-2012.

In the worst case, we can use the `1.2.643.2.2.9` OID for `Gost94Core<CryptoProParam>`.

I'm waiting for @IvashchenkoSerhii  to sched the light on `1.2.804.2.1.1.1.1.2.1` usage.

newpavlov: Sigh... The situation is quite annoying. We generally use OIDs for a "complete" function. It's not clear how we can support a function which has OID in a parametrized form and separate OIDs for its parameters.

>The problem is that the GOST 34.311-95 is referenced using the 1.2.643.2.2.9 OID rather than individual params OIDs. See, for example, the RFC4490 and its examples. Or likewise the examples from MR 26.2.002-2012.

IIUC in the RFC 4490 case the OID is given for the CryptoPro paramset:
>This function is always used with default parameters id-GostR3411-94-CryptoProParamSet (see Section 8.2 of [[CPALGS](https://www.rfc-editor.org/rfc/rfc4490#ref-CPALGS)]).

Maybe MR 26.2.002-2012 is the same?

lumag:
> I wonder how the [retracted TC26 OID](https://tc26.ru/about/protsedury-i-reglamenty/identifikatory-obektov-oid-tekhnicheskogo-komiteta-po-standartizatsii-kriptograficheskaya-zashchita-1.html) `1.2.643.7.1.1.2.1` relates to the ones used in this PR. IIUC the OIDs are introduced by the CryptoPro company and it's not clear how "official" they are.

If I understand correctly this started in the same way as the story of RIPEMD OIDs. However unlike RIPEMD, the TC26 retracted duplicating OIDs. 

Now back to usage question. The OID `1.2.643.2.2.30.1` was actively used in the public key params. The OID `1.2.643.2.2.9` was used in the signed CMS files to denote hashing algorithm. I haven't seen retracted TC26 OIDs in wild.

newpavlov:
For now I think we should use `1.2.643.2.2.9` for `Gost94Core<CryptoProParam>` with a reference to RFC 4490 and remove the impl for `TestParams` (either way, test paramset should not be used in the wild). @IvashchenkoSerhii  or someone else may add impl for the DSTU params later if such need will arise.

It may be worth to add a comment to the crate docs about `1.2.643.2.2.30.1` vs `1.2.643.2.2.9`.

gost94/src/gost94_core.rs
lumag: Yes, the mentioned MR document doesn't specify any other params (or a way to specify them), so only CryptoPro params can be used with `1.2.643.2.2.9`.

lumag:
Done

gost94/src/params.rs
```diff
@@ -31,6 +34,14 @@ impl Gost94Params for CryptoProParam {
     const NAME: &'static str = "Gost94CryptoPro";
 }
 
+#[cfg(feature = "oid")]
+impl AssociatedOid for CryptoProParam {
+    /// Per the RFC 4490, this OID is used for the GOST R 34.11-94 hash with CryptoPro params.
+    /// The OID 1.2.643.2.2.30.1 is used in the PublicKey params to denote CryptoPro paramset, but
+    /// not the hash function itself.
```
newpavlov: According to this description we should use:
```rust
impl AssociatedOid for Gost94Core<CryptoProParam> {
    const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.643.2.2.9");
}

impl AssociatedOid for CryptoProParam {
    const OID: ObjectIdentifier = ObjectIdentifier::new_unwrap("1.2.643.2.2.30.1");
}
```

Of course, the blanket impl would be removed with code like this. It does create a certain ambiguity, but I think it's the best we can do.

Also comment like this probably should go the crate docs, otherwise it will be hard for users to discover it.

lumag: Done

newpavlov:
Thank you!

IvashchenkoSerhii:
OID `1.2.804.2.1.1.1.1.2.1` defined in this law https://zakon.rada.gov.ua/laws/show/z1399-12#Text.

It really is a GOST 34.311-95 with the DSTU sBox (https://zakon.rada.gov.ua/laws/show/z0729-07#Text search by `ДКЕ N 1`)

lumag:
@IvashchenkoSerhii thank you!

PR #422, opened 2022-10-11
sha2: update dependency `cpufeature` to 0.2.5

luojia65:
The previous versions have been yanked: https://github.com/RustCrypto/utils/pull/802


newpavlov:
I don't see why it needs to be done. Building from scratch or running `cargo update` would result in v0.2.5 or later to be used. Updating `sha2`'s Cargo.toml will not affect those who keep their `Cargo.lock`.

PR #432, opened 2022-11-10
Fix `KeyInit` for blake2

vlad9486:
Right now it is crashes always on `KeyInit::new`.

### DIFF
--- a/blake2/src/macros.rs
+++ b/blake2/src/macros.rs
@@ -325,7 +325,7 @@ macro_rules! blake2_mac_impl {
                 let mut padded_key = Block::<$hash>::default();
                 padded_key[..kl].copy_from_slice(key);
                 Self {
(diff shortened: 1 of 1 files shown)


newpavlov:
Your changes are incorrect, you try to pass key as salt. How exactly does it crash?

vlad9486:
@newpavlov 

Now it is crashing:

```
#[cfg(test)]
#[test]
fn blake2b_trivial() {
    use blake2::{Blake2bMac512, digest::KeyInit};

    <Blake2bMac512 as KeyInit>::new(&Default::default());
}
```

Because it is your code is incorrect and passing key as salt. My fix is fixing that. And the mentioned test is passing.

vlad9486:
@newpavlov 
Please, look again. The original code is incorrect and passing key as salt. My fix is fixing that. Here is the test, which is crashed #433.

newpavlov:
Oh... True. We have tests for the MAC types, but they different codepath.

PR #440, opened 2022-12-09
blake2: add Cargo feature to optimize for code size.

Dirbaio:
The implementation uses some `inline(always)` to force inline the `round` and `quarter_round` functions for performance. This PR adds an opt-in Cargo feature to disable them, to allow optimizing for size instead.

Building for `thumbv6m-none-eabi`: 
- `size_opt` disabled: 5772 bytes
- `size_opt` enabled: 2372 bytes

This saves 3.4kb of code size, which is 60%.

I'm trying to use blake2 on a microcontroller with 32kb of Flash memory (where the code is stored), so saving 3.4kb is quite substantial.

### DIFF
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -25,3 +25,4 @@ reset = [] # Enable reset functionality
 simd = []
 simd_opt = ["simd"]
 simd_asm = ["simd_opt"]
+size_opt = [] # Optimize for code size. Removes some `inline(always)`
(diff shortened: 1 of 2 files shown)


blake2/Cargo.toml
```diff
@@ -25,3 +25,4 @@ reset = [] # Enable reset functionality
 simd = []
 simd_opt = ["simd"]
 simd_asm = ["simd_opt"]
+size_opt = [] # Optimize for code size. Removes some `inline(always)`
```
newpavlov: Could you add a new line here?

Dirbaio: Oops, done!

newpavlov:
Thank you! I will release it shortly.

Issue #315, opened 2021-09-09
Disable ASM hashes on MSVC?


not-an-aardvark:
As documented in https://github.com/RustCrypto/asm-hashes/issues/17 and elsewhere, the crates in `asm-hashes` don't work on Windows MSVC targets. This is unfortunate, but it's understandable given that assembly is inherently difficult to port to all platforms.

However, it seems unnecessary that this detail leaks into the crates in the `hashes` repository. Would it make sense to use conditional compilation to force the software versions of hashes on MSVC, even when the `asm` feature is enabled? (There is [already precedent](https://github.com/RustCrypto/hashes/blob/726e3c3978a77eb4318b4511eed5ac885375acd0/sha1/src/compress.rs#L23-L26) for ignoring the `asm` feature flag when no assembly builds are available for the current platform.)

My use case is that I want to enable the `asm` feature to allow for faster performance on compatible platforms, but I don't want to completely break the build for Windows users. I'd be happy to create a PR if you think this is a reasonable change.

newpavlov:
Sounds reasonable. But I wonder whether we should use black- or white-list approach here (i.e. disable for MSVC or enable for GNU and other supported toolchains).

newpavlov:
Dependency on asm-hashes will be removed in v0.11 releases.

tarcieri:
@newpavlov should we close the remaining issues on https://github.com/RustCrypto/asm-hashes and archive it?

newpavlov:
Yes, but I think we should do it after v0.11 hashes are released.

PR #447, opened 2023-01-16
md5: add inline assembly support for `x86_64` and `x86`

johnmave126:
This PR implements md5 using inline assembly, which has the benefit of compiling and running on MSVC targets.

This serves as a demo to a path to address #315, https://github.com/RustCrypto/asm-hashes/issues/45, and https://github.com/RustCrypto/asm-hashes/issues/17

## Performance Consideration
In the assembly implementation, a 3-way addition (2 registers + 1 immediate) is implemented as
```asm
lea <dst>, [<dst> + <index> + <displacement>]
```
https://github.com/johnmave126/hashes/blob/40a315dea9d5619e664c3b70749d0af692bca8b8/md5/src/asm/x86.rs#L81

However in Intel processors before Ice Lake, 3-way `LEA` has a latency of 3 cycles and uses only 1 port. (see [here](https://uops.info/html-instr/LEA_B_I_D32_R32.html#CLX)) In contrast, `LEA` has a latency of 1 and can be dispatched in parallel to 2 ports on newer processors.

In older processors, the following is better:
```asm
lea <dst>, [<dst> + <displacement>]
add <dst>, <index>
```

LLVM always emits something similar to the second flavor. Hence inline assembly will perform slower on processors before Ice Lake, but faster afterwards.

On Intel i7-6700K (Skylake), ubuntu-22.04.1:
```sh
> cargo +nightly bench
running 4 tests
test md5_10    ... bench:          15 ns/iter (+/- 2) = 666 MB/s
test md5_100   ... bench:         143 ns/iter (+/- 16) = 699 MB/s
test md5_1000  ... bench:       1,453 ns/iter (+/- 16) = 688 MB/s
test md5_10000 ... bench:      14,162 ns/iter (+/- 27) = 706 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 0.87s
```
```sh
> cargo +nightly bench --features inline-asm
running 4 tests
test md5_10    ... bench:          16 ns/iter (+/- 0) = 625 MB/s
test md5_100   ... bench:         162 ns/iter (+/- 0) = 617 MB/s
test md5_1000  ... bench:       1,661 ns/iter (+/- 2) = 602 MB/s
test md5_10000 ... bench:      16,307 ns/iter (+/- 14) = 613 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 1.16s
```
Using `LEA+ADD` in inline assembly still beats non-asm on older processors, but not by much:
```sh
> cargo +nightly bench --features inline-asm
running 4 tests
test md5_10    ... bench:          14 ns/iter (+/- 0) = 714 MB/s
test md5_100   ... bench:         145 ns/iter (+/- 0) = 689 MB/s
test md5_1000  ... bench:       1,412 ns/iter (+/- 6) = 708 MB/s
test md5_10000 ... bench:      13,787 ns/iter (+/- 16) = 725 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 0.98s
```
---

On AMD Ryzen 9 5900X (Zen 3), Windows 10 22H2:
```sh
> cargo +nightly bench
running 4 tests
test md5_10    ... bench:          13 ns/iter (+/- 0) = 769 MB/s
test md5_100   ... bench:         124 ns/iter (+/- 4) = 806 MB/s
test md5_1000  ... bench:       1,227 ns/iter (+/- 17) = 814 MB/s
test md5_10000 ... bench:      12,255 ns/iter (+/- 136) = 815 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 1.88s
```
```sh
> cargo +nightly bench --features inline-asm
running 4 tests
test md5_10    ... bench:          11 ns/iter (+/- 0) = 909 MB/s
test md5_100   ... bench:         102 ns/iter (+/- 1) = 980 MB/s
test md5_1000  ... bench:       1,023 ns/iter (+/- 10) = 977 MB/s
test md5_10000 ... bench:      10,434 ns/iter (+/- 366) = 958 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 6.79s
```

## Dependency
The inline assembly code for md5 comes from a side project by me, which uses a declarative macro crate ([asm_block](https://crates.io/crates/asm_block) also by me) to help defining macros emitting assembly. This introduces an additional dependency. If that's unwanted, we can strip down the `asm_block` macro to minimally viable and embed it directly.

## MSRV
Enabling inline assembly will bump MSRV to **1.59**.

## CI
The PR also includes an additional CI job to test this feature.

### DIFF
--- a/.github/workflows/md5.yml
+++ b/.github/workflows/md5.yml
@@ -83,3 +83,20 @@ jobs:
           toolchain: ${{ matrix.rust }}
           override: true
       - run: cargo test --features oid
+
+  # TODO: merge with test on MSRV bump to 1.59 or higher
+  test-inline-asm:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - 1.59.0 # MSRV
+    steps:
+      - uses: actions/checkout@v3
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: actions-rs/toolchain@v1
+        with:
+          profile: minimal
+          toolchain: ${{ matrix.rust }}
+          override: true
+      - run: cargo test --features inline-asm
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,12 @@
 version = 3
 
 [[package]]
+name = "asm_block"
+version = "0.1.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "466c0990cf15ef0f331f19fdc16fd60229606ab237476c70a66b747fce2911ab"
+
+[[package]]
 name = "blake2"
 version = "0.10.6"
 dependencies = [
@@ -156,6 +162,7 @@ checksum = "db6d7e329c562c5dfab7a46a2afabc8b987ab9a4834c9d1ca04dc54c1546cef8"
 name = "md-5"
 version = "0.10.5"
 dependencies = [
+ "asm_block",
  "digest",
  "hex-literal",
  "md5-asm",
--- a/md5/Cargo.toml
+++ b/md5/Cargo.toml
@@ -19,6 +19,7 @@ digest = "0.10.4"
 
 [target.'cfg(any(target_arch = "x86", target_arch = "x86_64"))'.dependencies]
 md5-asm = { version = "0.5", optional = true }
+asm_block = { version = "0.1.3", optional = true }
 
 [dev-dependencies]
 digest = { version = "0.10.4", features = ["dev"] }
@@ -28,4 +29,5 @@ hex-literal = "0.2.2"
 default = ["std"]
 std = ["digest/std"]
 asm = ["md5-asm"] # WARNING: this feature SHOULD NOT be enabled by library crates
+inline-asm = ["asm_block"] # Enable inline assembly support. WARNING: Bumps MSRV to 1.59
 oid = ["digest/oid"] # Enable OID support. WARNING: Bumps MSRV to 1.57
--- a/md5/README.md
+++ b/md5/README.md
@@ -28,6 +28,8 @@ including HMAC-MD5.
 
 Rust **1.41** or higher.
 
+Enabling feature flag `inline-asm` requires Rust **1.59** or higher.
+
 Minimum supported Rust version can be changed in the future, but it will be
 done with a minor version bump.
 
--- a/md5/src/asm/mod.rs
+++ b/md5/src/asm/mod.rs
@@ -1,0 +1,12 @@
+#[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
+mod x86;
+
+#[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
+pub use x86::compress_block;
+
+#[inline]
+pub fn compress(state: &mut [u32; 4], blocks: &[[u8; 64]]) {
+    for block in blocks {
+        compress_block(state, block)
+    }
+}
--- a/md5/src/asm/mod.rs
+++ b/md5/src/asm/mod.rs
@@ -2,11 +2,4 @@
 mod x86;
 
 #[cfg(any(target_arch = "x86_64", target_arch = "x86"))]
-pub use x86::compress_block;
-
-#[inline]
-pub fn compress(state: &mut [u32; 4], blocks: &[[u8; 64]]) {
-    for block in blocks {
-        compress_block(state, block)
-    }
-}
+pub use x86::compress;
--- a/md5/src/asm/x86.rs
+++ b/md5/src/asm/x86.rs
@@ -1,0 +1,280 @@
+//! MD5 assembly code for `x86_64` and `x86`. Adapted from Project Nayuki.
+/*
+ * MD5 hash in x86-64 assembly
+ *
+ * Copyright (c) 2016 Project Nayuki. (MIT License)
+ * https://www.nayuki.io/page/fast-md5-hash-implementation-in-x86-assembly
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy of
+ * this software and associated documentation files (the "Software"), to deal in
+ * the Software without restriction, including without limitation the rights to
+ * use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+ * the Software, and to permit persons to whom the Software is furnished to do so,
+ * subject to the following conditions:
+ * - The above copyright notice and this permission notice shall be included in
+ *   all copies or substantial portions of the Software.
+ * - The Software is provided "as is", without warranty of any kind, express or
+ *   implied, including but not limited to the warranties of merchantability,
+ *   fitness for a particular purpose and noninfringement. In no event shall the
+ *   authors or copyright holders be liable for any claim, damages or other
+ *   liability, whether in an action of contract, tort or otherwise, arising from,
+ *   out of or in connection with the Software or the use or other dealings in the
+ *   Software.
+ */
+use core::arch::asm;
+
+use asm_block::asm_block;
+
+/// MD5 operators
+macro_rules! asm_md5_op {
+    (F, $a: tt, $b: tt, $c: tt, $d: tt, $k: tt, $s: literal, $t: literal, $tmp1: tt, $tmp2: tt) => {
+        concat!(
+            asm_block! {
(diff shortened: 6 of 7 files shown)


newpavlov:
Thank you!

After a cursory look, I have the following comments:
- It could be worth to move the block iteration loop into the asm block. It could improve performance a bit since the hasher state will be kept in registers without spilling into memory. LLVM is not always able to move loads and stores outside of a loop, so it's better to be explicit.
- I thought the RBX register is reserved by LLVM and should not be used by inline asm. Are you sure your code is correct in this regard?
- What about binary code size? Could be the performance difference be explained by aggressive inlining?
- How much worse the code would be without `asm_block`? We are somewhat conservative about third-party dependencies, but your macro looks quite useful for repetitive assembly and we may use it in other crates in future. Would you be willing to transfer it to the [utils](https://github.com/RustCrypto/utils) repo?

I will try to do a more thorough review somewhat later.

johnmave126:
> It could be worth to move the block iteration loop into the asm block. It could improve performance a bit since the hasher state will be kept in registers without spilling into memory. LLVM is not always able to move loads and stores outside of a loop, so it's better to be explicit.

Worth a try. Will try that during the weekend. The number of registers on `x86_64` supports this, but I haven't given it a thorough thought on how to do this on `x86` yet.

> I thought the RBX register is reserved by LLVM and should not be used by inline asm. Are you sure your code is correct in this regard?

`sp` is always reserved. `bp` and `si` are only reserved on `x86` and `bx` is only reserved on `x86_64`.

> What about binary code size? Could be the performance difference be explained by aggressive inlining?

binary size for bench executable without feature
```
> du -s ../target/release/deps/mod-bbe1efd848abf6c7
6532    ../target/release/deps/mod-bbe1efd848abf6c7
```

binary size for bench executable with feature `inline-asm`
```
> du -s ../target/release/deps/mod-2aedef2c10a10821
6532    ../target/release/deps/mod-2aedef2c10a10821
```
Not sure why they are the same. Did I do something wrong?

I inspected the assembly using `cargo asm`, both versions are inlined very similarly. The main reason I think it was due to `lea` is that assuming `add+add` on `x86`, the approximate number of cycles for the md5 assembly is 592. Changing it to `lea` adds 1 cycle to each md5 operator, making it now 656 cycles. `592/656` is approximately 0.9, which is very close to the observed performance difference.

> How much worse the code would be without `asm_block`? We are somewhat conservative about third-party dependencies, but your macro looks quite useful for repetitive assembly and we may use it in other crates in future. Would you be willing to transfer it to the [utils](https://github.com/RustCrypto/utils) repo?

If we don't introduce any utility, the best we can do would be something like this:
```rust
#[cfg_attr(rustfmt, rustfmt_skip)]
macro_rules! op_f {
    ($a: literal, $b: literal, $c: literal, $d: literal, $t: literal, $s: literal, $k: literal, $tmp: literal) => {
        concat!(
            "mov ", $tmp, ", ",  $c, "\n",
            "add ", $a, ", ", $t, "\n",
            "xor ", $tmp, ", ", $d, "\n",
            "and ", $tmp, ", ", $b, "\n",
            "xor ", $tmp, ", ", $d, "\n",
            "lea ", $a, ", [", $tmp, " + ", $a, " + ", $k, "]\n",
            "rol ", $a, ", ", $s, "\n",
            "add ", $a, ", ", $b, "\n",
        )
    };
}
```
And call site will be something like
```rust
op_f!("{a:e}", "{b:e}", "{c:e}", "{d:e}", "[{x} + 0]", 7, 0xd76aa478, "{t1:e}")
```
I personally find the concatenation is hard to read, especially when there are commas in the string. The main disadvantage is that it was even harder to write.

And I don't mind moving my crate elsewhere. Just let me know how it should be done.

johnmave126:
Tested on AMD Ryzen 9 5900X (Zen 3), Windows 10 22H2:

## Before moving block loop inside asm
`x86_64`:
```
running 4 tests
test md5_10    ... bench:          11 ns/iter (+/- 0) = 909 MB/s
test md5_100   ... bench:         102 ns/iter (+/- 5) = 980 MB/s
test md5_1000  ... bench:       1,016 ns/iter (+/- 31) = 984 MB/s
test md5_10000 ... bench:      10,190 ns/iter (+/- 302) = 981 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 2.99s
```
`x86`:
```
running 4 tests
test md5_10    ... bench:          11 ns/iter (+/- 0) = 909 MB/s
test md5_100   ... bench:         104 ns/iter (+/- 4) = 961 MB/s
test md5_1000  ... bench:       1,017 ns/iter (+/- 13) = 983 MB/s
test md5_10000 ... bench:      10,189 ns/iter (+/- 235) = 981 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 4.28s
```

## After moving block loop inside asm
`x86_64`:
```
running 4 tests
test md5_10    ... bench:          11 ns/iter (+/- 0) = 909 MB/s
test md5_100   ... bench:         102 ns/iter (+/- 2) = 980 MB/s
test md5_1000  ... bench:       1,002 ns/iter (+/- 17) = 998 MB/s
test md5_10000 ... bench:       9,959 ns/iter (+/- 72) = 1004 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 2.22s
```
`x86`:
```
running 4 tests
test md5_10    ... bench:          11 ns/iter (+/- 0) = 909 MB/s
test md5_100   ... bench:         103 ns/iter (+/- 1) = 970 MB/s
test md5_1000  ... bench:       1,002 ns/iter (+/- 38) = 998 MB/s
test md5_10000 ... bench:       9,931 ns/iter (+/- 112) = 1006 MB/s

test result: ok. 0 passed; 0 failed; 0 ignored; 4 measured; 0 filtered out; finished in 1.94s
```

oech3:
Any progress? This PR is conflicting.

oech3:
@sjthomason Are you interested in adding same optimization with https://github.com/ogital-net/fast-md5 to this crate?

I think the goal of this PR is near with `fast-md5`, but this PR is inactive for a long time.

tarcieri:
Note that we'd still be happy to have this PR landed here, however it has conflicts that need to be resolved

sjthomason:
Instead of trying to rebase this one, I opened #917 off of latest master.  Let me know if you'd prefer to continue the work here.

Thanks!

tarcieri:
@sjthomason thanks! I'll close this in favor of #917

oech3:
I tried #917 at uutils's `cksum -a md5`. I got
```
> hyperfine -N "target/release/cksum -a md5 /lib/chromium/chromium" "target/release/cksum-openssl -a md5 /lib/chromium/chromium" "target/release/cksum-pr917 -a md5 /lib/chromium/chromium"
Benchmark 1: target/release/cksum -a md5 /lib/chromium/chromium
  Time (mean ± σ):     458.4 ms ±   0.9 ms    [User: 438.1 ms, System: 19.0 ms]
  Range (min … max):   457.2 ms … 460.2 ms    10 runs
 
Benchmark 2: target/release/cksum-openssl -a md5 /lib/chromium/chromium
  Time (mean ± σ):     368.5 ms ±   0.3 ms    [User: 345.6 ms, System: 21.8 ms]
  Range (min … max):   368.1 ms … 369.2 ms    10 runs
 
Benchmark 3: target/release/cksum-pr917 -a md5 /lib/chromium/chromium
  Time (mean ± σ):     367.6 ms ±   0.4 ms    [User: 347.7 ms, System: 18.8 ms]
  Range (min … max):   366.9 ms … 368.1 ms    10 runs
 
Summary
  target/release/cksum-pr917 -a md5 /lib/chromium/chromium ran
    1.00 ± 0.00 times faster than target/release/cksum-openssl -a md5 /lib/chromium/chromium
    1.25 ± 0.00 times faster than target/release/cksum -a md5 /lib/chromium/chromium
```
It would be nice if it was included to next version of uutils.

PR #454, opened 2023-03-02
Replace unmaintained actions-rs/toolchain action in CI workflows

striezel:
Basically all of the `actions-rs/*` actions are unmaintained. See <https://github.com/actions-rs/toolchain/issues/216> for more information. Due to their age they generate several warnings in CI runs, for example in https://github.com/RustCrypto/hashes/actions/runs/3716377913:

> Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: actions-rs/toolchain@v1

To get rid of some of those warnings the occurrences of `actions-rs/toolchain` are replaced by [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain).

### DIFF
--- a/.github/workflows/belt-hash.yml
+++ b/.github/workflows/belt-hash.yml
@@ -31,12 +31,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: RustCrypto/actions/cargo-cache@master
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          target: ${{ matrix.target }}
-          override: true
+          targets: ${{ matrix.target }}
       - run: cargo build --no-default-features --target ${{ matrix.target }}
 
   minimal-versions:
@@ -54,11 +52,9 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: RustCrypto/actions/cargo-cache@master
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          override: true
       - run: cargo check --all-features
       - run: cargo test --no-default-features
       - run: cargo test
--- a/.github/workflows/blake2.yml
+++ b/.github/workflows/blake2.yml
@@ -31,12 +31,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: RustCrypto/actions/cargo-cache@master
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
-          target: ${{ matrix.target }}
-          override: true
+          targets: ${{ matrix.target }}
       - run: cargo build --no-default-features --target ${{ matrix.target }}
 
   minimal-versions:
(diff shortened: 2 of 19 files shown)


.github/workflows/belt-hash.yml
```diff
@@ -31,12 +31,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: RustCrypto/actions/cargo-cache@master
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
         with:
-          profile: minimal
           toolchain: ${{ matrix.rust }}
           target: ${{ matrix.target }}
```
tarcieri: These need to be changed to `targets:`

```suggestion
          targets: ${{ matrix.target }}
```

See: https://github.com/dtolnay/rust-toolchain#inputs

tarcieri:
Thanks. We've been making this change elsewhere (e.g. https://github.com/RustCrypto/formats/pull/887) but there are quite a few repos to update

.github/workflows/belt-hash.yml
striezel: Yeah, the documentation of `dtolnay/rust-toolchain` says so, but when you look into the [action.yml](https://github.com/dtolnay/rust-toolchain/blob/e12eda571dc9a5ee5d58eecf4738ec291c66f295/action.yml) it also mentions `target` as an alias for `targets`. So both variants should work, but I'll change it anyway.

tarcieri: Aah, interesting

striezel: I guess that alias was intentionally added for easier migration from `actions-rs/toolchain` where the parameter is named `target`.

tarcieri (approved):

tarcieri:
Hmm, I think GitHub Actions is having some issues... jobs are showing up as cancelled but without failures

tarcieri:
@striezel can you `commit --amend` and try re-pushing? CI seems wedged

.github/workflows/belt-hash.yml
```diff
@@ -31,12 +31,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: RustCrypto/actions/cargo-cache@master
-      - uses: actions-rs/toolchain@v1
+      - uses: dtolnay/rust-toolchain@master
```
newpavlov, line 34: Is there a reason for using `@master` instead of `@v1`?

striezel:
> @striezel can you `commit --amend` and try re-pushing? CI seems wedged

Done and force pushed. And while I was at it, I also replaced the remaining `target: ...` with `targets: ...` for consistency.

.github/workflows/belt-hash.yml
striezel: No particular reason, just force of habit. I basically just followed the example in the readme of `dtolnay/toolchain` which uses `@master`, too: https://github.com/dtolnay/rust-toolchain/tree/e12eda571dc9a5ee5d58eecf4738ec291c66f295#toolchain-expressions
Do you want me to replace it with `@v1` instead?

newpavlov: Ah, it looks like `@v1` is not properly supported by the action: https://github.com/dtolnay/rust-toolchain/issues/45

newpavlov:
Thank you!

PR #457, opened 2023-03-12
Add CI `no_std` optional features integration check

pinkforest:
Tests all `no_std` features with cargo hack in a target that does not have `std` available

Some optional features may break `no_std` integration that may accidentally bring `std`

Mirrors what we are doing for dalek:
- https://github.com/dalek-cryptography/curve25519-dalek/pull/512

### DIFF
--- a/.github/workflows/belt-hash.yml
+++ b/.github/workflows/belt-hash.yml
@@ -59,3 +59,17 @@ jobs:
       - run: cargo test --no-default-features
       - run: cargo test
       - run: cargo test --all-features
+
+  # Build iterate all no_std feature combos
+  build-nostd-features:
+    name: Build features on no_std target (thumbv7em-none-eabi)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


newpavlov:
The blake2 job fails because it tries to test Nightly-only features on stable channel.

pinkforest:
Yeah I have a commit ready in push but I didn't want to burn CI before I confirm couple of things re: toolchain in zulip

As the toolchain on it seems to be pinned on some specific nightly I was wondering about

.github/workflows/belt-hash.yml
```diff
@@ -59,3 +59,17 @@ jobs:
       - run: cargo test --no-default-features
       - run: cargo test
       - run: cargo test --all-features
+
+  # Build iterate all no_std feature combos
+  build-nostd-features:
```
tarcieri, line 64: I think you could probably put this under the build job (which is probably more accurately named `no_std` or `build-no_std`), which would reuse intermediate artifacts

newpavlov:
Closing in favor of #485

PR #467, opened 2023-04-17
Fix an off-by-one error in the docs for sha2

carols10cents:
👋🏻 Just a little drive-by docs fix!

There are six total, and the docs mentioned "first two" and "last three" = 5.

This appears to have been introduced in 75298d3, and based on the text in this spot that was removed in that commit, this should read "last four" instead.


tarcieri:
Thank you!

PR #470, opened 2023-04-26
Empty OID crashes the cargo doumentation generation with oid feature

mrnossiom:
👋🏻 

While building a project with the `pgp` lib, I could not build the documentation due to failed evaluation of constant value. Such empty OID generations would make `const-oid` panic due to how `ObjectIdentifier::new_unwrap` is made. I don't think this is removing any feature, since it would panic during runtime.

I took the opportunity to fix a rustdoc lint.


tarcieri (approved):
Looks fine... can you remove draft?

mrnossiom:
For sure

PR #471, opened 2023-04-30
chore: fix link to `ascon-hash` crate

TomAFrench:
The link to the `ascon-hash` crate was broken due to missing a reference. I've added this reference.


tarcieri:
Thank you!

PR #353, opened 2022-01-23
Refactor k12

aewag:
The k12 crate had its own keccak-p permutation implementation. 
~With https://github.com/RustCrypto/sponges/pull/7 a generic keccak-p implementation would be available, which could be used within the k12 crate.~
With https://github.com/RustCrypto/hashes/pull/458 TurboSHAKE is available, which can be used within the `k12` crate.

`k12` uses a block size of 8192 to split the input. The maximum supported block size of `block-buffer` is limited to 256. Therefore this implementation buffers the input within an additional array.

PR #475, opened 2023-05-08
Draft: Add MarsupilamiFourteen (m14)

aewag:
This is a draft to add the `MarsupilamiFourteen` hash function (related issue #1).
As `m14` is a variant of `KangarooTwelve`, the implementation can reuse parts of the `k12` crate.
(Ignore the first commit as it is part of the open PR #353)

Therefore, the `k12` crate is modified to export a macro to generate the hasher. Further is the `sha3` crate changed to expose a more low-level, but "hazmat", API.
Shall this gated behind a "hazmat" feature?

~I didnot find yet any test vectors, therefore tests are currently missing, but will be added.~


gvanas:
Of course, nothing prevents you from adding M14 to RustCrypto. But as written [here](https://github.com/XKCP/XKCP/pull/115), I would have a 256-bit secure function, similar to K12, that calls TurboSHAKE256 instead of TurboSHAKE128. Given the status on cryptanalysis, I don't see the need for the 14 rounds anymore; having 12 rounds everywhere just seems simpler and cleaner.

aewag:
> Of course, nothing prevents you from adding M14 to RustCrypto. But as written [here](https://github.com/XKCP/XKCP/pull/115), I would have a 256-bit secure function, similar to K12, that calls TurboSHAKE256 instead of TurboSHAKE128. Given the status on cryptanalysis, I don't see the need for the 14 rounds anymore; having 12 rounds everywhere just seems simpler and cleaner.

That definitely simplifies the implementation. I will therefore close this PR and better integrate a `k12-256` or similar to the `k12` crate.

PR #479, opened 2023-05-19
sha3: Add derive Zeroize for Sha3State

aewag:
Not zeroizing the Sha3State allows to recover any squeezed output. This is because the `keccak` permutations can be inversed. Hence, access to the complete state allows to perform this operation.

While this is security-relevant, including it would significantly increase the MSRV. Therefore, it is gated behind the `zeroize` feature.

@tarcieri what do you think about gating the zeroizing behind a feature?

The `ascon` implementation would "require" zeroizing as well. As the implementations differ, zeroizing has to be done in the `sponges` crate. Therefore, I would prepare a similar PR for the `sponges` repository targeting `ascon`.

### DIFF
--- a/sha3/src/state.rs
+++ b/sha3/src/state.rs
@@ -1,4 +1,6 @@
 use core::convert::TryInto;
+#[cfg(feature = "zeroize")]
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
 const PLEN: usize = 25;
 const DEFAULT_ROUND_COUNT: usize = 24;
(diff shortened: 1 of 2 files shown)


tarcieri:
Yes, behind a feature sounds good. If you avoid using the derive macro and write the impl by hand (which seems simple enough in this case), you can avoid the `syn` dependency as well.

aewag:
@tarcieri Thanks for the feedback. I updated the PR and added the `impl`. For documentation purpose: I only zeroed the `state` and not the `round_count`, as the `round_count` is not security-relevant.

sha3/src/state.rs
```diff
@@ -18,6 +20,20 @@ impl Default for Sha3State {
...
+        self.state.zeroize();
+    }
+}
+
+#[cfg(feature = "zeroize")]
+impl Drop for Sha3State {
+    fn drop(&mut self) {
+        self.zeroize();
+    }
+}
```
tarcieri: I would suggest either impl'ing the `Zeroize` trait *or* `Drop`+`ZeroizeOnDrop`.

aewag: I updated the PR with an impl of `Drop` + `ZeroizeOnDrop`.

newpavlov (approved):
In future we probably also should add `zeroize` feature to `digest`, which would depend on https://github.com/RustCrypto/utils/pull/832, but we can do it in later PRs.

tarcieri (approved):

PR #480, opened 2023-05-19
ascon-hash: Add Zeroize feature

aewag:
Not zeroizing the state allows to recover any squeezed output. This is because the `ascon` permutations can be inversed. Hence, access to the complete state allows to perform this operation.

This relies on https://github.com/RustCrypto/sponges/pull/57 and a new release of the `ascon` crate.

### DIFF
--- a/ascon-hash/Cargo.toml
+++ b/ascon-hash/Cargo.toml
@@ -19,7 +19,7 @@ rust-version = "1.56"
 
 [dependencies]
 digest = { version = "0.10", default-features = false, features = ["core-api"] }
-ascon = "0.3"
+ascon = { version = "0.4", default-features = false }
 
 [dev-dependencies]
 spectral = { version = "0.6", default-features = false }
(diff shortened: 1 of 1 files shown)


ascon-hash/Cargo.toml
```diff
@@ -19,15 +19,16 @@ rust-version = "1.56"
...
-ascon = "0.3"
+ascon = { version = "0.4", default-features = false }
 
 [dev-dependencies]
 spectral = { version = "0.6", default-features = false }
 hex = "0.4"
 
 [features]
-default = ["std"]
+default = ["std", "zeroize"]
```
newpavlov: Usually we do not enable `zeroize` by default.

aewag: Okay, in this case I will remove it :+1:

ascon-hash/Cargo.toml
```diff
@@ -19,15 +19,16 @@ rust-version = "1.56"
...
 std = ["digest/std"]
+zeroize = ["ascon/zeroize"]
```
newpavlov: It may be worth to add a TODO comment to remind to enable `zeroize` feature for `block-buffer` which will be available in v0.11. See https://github.com/RustCrypto/utils/pull/963

aewag: I updated this line

PR #484, opened 2023-06-08
JH hash implementation

vorot93:


jh/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+documentation = "https://docs.rs/jh"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["crypto", "jh", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+edition = "2021"
+rust-version = "1.57"
+
+[dependencies]
+block-buffer = "0.10"
+block-padding = "0.3"
```
newpavlov: I don't think you use `block-padding` anywhere in the code. `block-buffer` can be imported from the `digest` crate.

jh/src/consts.rs
```diff
@@ -0,0 +1,21 @@
+use hex_literal::hex;
+
+pub const JH224_H0: [u8; 128] = hex!("
+    2dfedd62f99a98acae7cacd619d634e7a4831005bc301216b86038c6c966149466d9899f2580706fce9ea31b1d9b1adc11e8325f7b366e10f994857f02fa06c1
+    1b4f1b5cd8c840b397f6a17f6e738099dcdf93a5adeaa3d3a431e8dec9539a6822b4a98aec86a1e4d574ac959ce56cf015960deab5ab2bbf9611dcf0dd64ea6e
```
newpavlov: It's better to write this and other constants as:
```rust
pub const JH224_H0: [u8; 128] = hex!(
    "2dfedd62f99a98acae7cacd619d634e7a4831005bc301216b86038c6c9661494"
    "66d9899f2580706fce9ea31b1d9b1adc11e8325f7b366e10f994857f02fa06c1"
    "1b4f1b5cd8c840b397f6a17f6e738099dcdf93a5adeaa3d3a431e8dec9539a68"
    "22b4a98aec86a1e4d574ac959ce56cf015960deab5ab2bbf9611dcf0dd64ea6e"
);
```

jh/src/lib.rs
```diff
@@ -0,0 +1,152 @@
...
+    block_buffer::Eager,
+    core_api::{
+        AlgorithmName, Block, Buffer, BufferKindUser, CoreWrapper, FixedOutputCore, UpdateCore,
+    },
+    crypto_common::{BlockSizeUser, OutputSizeUser},
+    generic_array::typenum::{Unsigned, U28, U32, U48, U64},
+    HashMarker, Output,
+};
+
+macro_rules! define_hasher {
```
newpavlov: Instead of this macro you can implement `VariableOutputCore` for `JhCore` and then define the variants as:
```rust
pub type Jh256 = CoreWrapper<CtVariableCoreWrapper<JhCore, U32>>;
```
For reference you can take a look at the [`sha2`](https://github.com/RustCrypto/hashes/blob/master/sha2/src/core_api.rs) crate.

PR #490, opened 2023-06-15
sha2: Add aarch64 backends for SHA2.

codahale:
Adds NEON-enabled backends for SHA2 on `aarch64`.

Eliminates the need for the `asm` feature on `aarch64` for SHA-{224, 256} performance and provides a big performance boost for SHA-512, which didn’t benefit from the `asm` feature.

Before:

```
test sha256_10    ... bench:          27 ns/iter (+/- 0) = 370 MB/s
test sha256_100   ... bench:         278 ns/iter (+/- 3) = 359 MB/s
test sha256_1000  ... bench:       2,747 ns/iter (+/- 24) = 364 MB/s
test sha256_10000 ... bench:      27,392 ns/iter (+/- 293) = 365 MB/s
test sha512_10    ... bench:          17 ns/iter (+/- 0) = 588 MB/s
test sha512_100   ... bench:         164 ns/iter (+/- 7) = 609 MB/s
test sha512_1000  ... bench:       1,650 ns/iter (+/- 28) = 606 MB/s
test sha512_10000 ... bench:      16,533 ns/iter (+/- 1,540) = 604 MB/s
```

After:

```
test sha256_10    ... bench:           4 ns/iter (+/- 0) = 2500 MB/s
test sha256_100   ... bench:          46 ns/iter (+/- 0) = 2173 MB/s
test sha256_1000  ... bench:         424 ns/iter (+/- 6) = 2358 MB/s
test sha256_10000 ... bench:       4,190 ns/iter (+/- 31) = 2386 MB/s
test sha512_10    ... bench:           6 ns/iter (+/- 0) = 1666 MB/s
test sha512_100   ... bench:          65 ns/iter (+/- 0) = 1538 MB/s
test sha512_1000  ... bench:         636 ns/iter (+/- 5) = 1572 MB/s
test sha512_10000 ... bench:       6,311 ns/iter (+/- 68) = 1584 MB/s
```

(Benchmarks run on my M2 Air laptop, unplugged, on my kitchen table.)

### DIFF
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -13,7 +13,7 @@ cfg_if::cfg_if! {
         }
         mod x86;
         use x86::compress;
-    } else if #[cfg(target_arch = "aarch64")] {
+    } else if #[cfg(all(feature = "asm", target_arch = "aarch64"))] {
         mod soft;
         mod aarch64;
         use aarch64::compress;
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -13,7 +13,7 @@ cfg_if::cfg_if! {
         }
         mod x86;
         use x86::compress;
-    } else if #[cfg(all(feature = "asm", target_arch = "aarch64"))] {
+    } else if #[cfg(target_arch = "aarch64")] {
         mod soft;
         mod aarch64;
         use aarch64::compress;
--- a/sha2/src/sha256/aarch64.rs
+++ b/sha2/src/sha256/aarch64.rs
@@ -112,7 +112,11 @@ unsafe fn vsha256hq_u32(
     hash_abcd: uint32x4_t,
     wk: uint32x4_t,
 ) -> uint32x4_t {
-    asm!("SHA256H {:q}, {:q}, {:v}.4S", inout(vreg) hash_efgh, in(vreg) hash_abcd, in(vreg) wk);
+    asm!(
+        "SHA256H {:q}, {:q}, {:v}.4S",
+        inout(vreg) hash_efgh, in(vreg) hash_abcd, in(vreg) wk,
+        options(pure, nomem, nostack, preserves_flags)
+    );
     hash_efgh
 }
 
@@ -122,13 +126,21 @@ unsafe fn vsha256h2q_u32(
     hash_abcd: uint32x4_t,
     wk: uint32x4_t,
 ) -> uint32x4_t {
-    asm!("SHA256H2 {:q}, {:q}, {:v}.4S", inout(vreg) hash_efgh, in(vreg) hash_abcd, in(vreg) wk);
+    asm!(
+        "SHA256H2 {:q}, {:q}, {:v}.4S",
+        inout(vreg) hash_efgh, in(vreg) hash_abcd, in(vreg) wk,
+        options(pure, nomem, nostack, preserves_flags)
+    );
     hash_efgh
 }
 
 #[inline(always)]
 unsafe fn vsha256su0q_u32(mut w0_3: uint32x4_t, w4_7: uint32x4_t) -> uint32x4_t {
-    asm!("SHA256SU0 {:v}.4S, {:v}.4S", inout(vreg) w0_3, in(vreg) w4_7);
+    asm!(
+        "SHA256SU0 {:v}.4S, {:v}.4S",
+        inout(vreg) w0_3, in(vreg) w4_7,
+        options(pure, nomem, nostack, preserves_flags)
+    );
     w0_3
 }
 
@@ -138,6 +150,10 @@ unsafe fn vsha256su1q_u32(
     w8_11: uint32x4_t,
     w12_15: uint32x4_t,
 ) -> uint32x4_t {
-    asm!("SHA256SU1 {:v}.4S, {:v}.4S, {:v}.4S", inout(vreg) tw0_3, in(vreg) w8_11, in(vreg) w12_15);
+    asm!(
+        "SHA256SU1 {:v}.4S, {:v}.4S, {:v}.4S",
+        inout(vreg) tw0_3, in(vreg) w8_11, in(vreg) w12_15,
+        options(pure, nomem, nostack, preserves_flags)
+    );
     tw0_3
 }
--- a/sha2/src/sha256/aarch64.rs
+++ b/sha2/src/sha256/aarch64.rs
@@ -1,6 +1,13 @@
 //! SHA-256 `aarch64` backend.
 
+// Implementation adapted from mbedtls.
+
 // TODO: stdarch intrinsics: RustCrypto/hashes#257
+
+use core::arch::aarch64::*;
+use core::arch::asm;
+
+use crate::consts::K32;
 
 cpufeatures::new!(sha2_hwcap, "sha2");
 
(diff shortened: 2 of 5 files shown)


tarcieri:
Neat! I hadn't thought about using inline ASM as a sort of "polyfill" for using unstable intrinsics on stable Rust before.

This is something we should consider doing elsewhere we use unstable aarch64 intrinsics, such as in the `aes` and `polyval` crates. Then, when the intrinsics are stabilized, we can delete the inline ASM and switch to the intrinsics.

sha2/src/sha256/aarch64.rs
```diff
@@ -1,15 +1,134 @@
...
+
+macro_rules! vsha256su1q_u32 {
+    ($tw0_3:expr, $w8_11:expr, $w12_15:expr) => {{
+        let mut out = $tw0_3;
+        asm!("SHA256SU1 {:v}.4S, {:v}.4S, {:v}.4S", inout(vreg) out, in(vreg) $w8_11, in(vreg) $w12_15);
+        out
+    }};
+}
+
+unsafe fn sha256_compress(state: &mut [u32; 8], block: &[u8; 64]) {
```
newpavlov: Add `SAFETY:` documentation with the list of required target features. Same for `sha512_compress`.

Also it may be useful to move iteration over blocks into this function to prevent redundant load and stores of the state.

newpavlov:
Update the cross version to fix the CI failures.

sha2/src/sha512/aarch64.rs
```diff
@@ -0,0 +1,210 @@
...
+    gh = vaddq_u64(gh, intermed);
+
+    // Rounds 14 and 15
+    initial_sum = vaddq_u64(s7, vld1q_u64(&K64[14]));
+    sum = vaddq_u64(vextq_u64(initial_sum, initial_sum, 1), ab);
+    intermed = vsha512hq_u64!(sum, vextq_u64(gh, ab, 1), vextq_u64(ef, gh, 1));
+    ab = vsha512h2q_u64!(intermed, ef, cd);
+    ef = vaddq_u64(ef, intermed);
+
+    for t in (16..80).step_by(16) {
```
newpavlov: Maybe it's worth to fully unroll this loop?

sha2/src/sha256/aarch64.rs
```diff
@@ -1,15 +1,134 @@
...
-        sha2_asm::compress256(state, blocks);
+        for block in blocks {
+            unsafe { sha256_compress(state, block) }
+        }
     } else {
         super::soft::compress(state, blocks);
     }
 }
+
+macro_rules! vsha256hq_u32 {
```
tarcieri: Is there some reason I'm missing why this is a macro and not just `unsafe fn vsha256hq_u32(hash_abcd: uint32x4_t, ...) -> uint32x4_t`?

Possibly with `#[inline(always)]` sprinkled on it.

sha2/src/sha256.rs
```diff
@@ -13,7 +13,7 @@ cfg_if::cfg_if! {
         }
         mod x86;
         use x86::compress;
-    } else if #[cfg(all(feature = "asm", target_arch = "aarch64"))] {
+    } else if #[cfg(target_arch = "aarch64")] {
```
newpavlov: We should keep the feature for now, since according to our policy MSRV bump is a breaking change. Add a TODO comment to remove the feature-based cfg in v0.11.

sha2/src/sha256/aarch64.rs
codahale: IME, Rust can’t fully inline functions in the same way it inlines intrinsics. It introduces some value copying which I haven’t been able to eliminate and which can have negative effects on performance relative to the macro approach. I suspect it’s due to `asm!` being a bit of a black box for the compiler but I’m a bit out of my depth there.

“This works gooder on my laptop” is probably the more concise explanation, though.

codahale:
> Update the cross version to fix the CI failures.

What should I bump it to?

newpavlov:
>What should I bump it to?

1.59

sha2/src/sha256/aarch64.rs
tarcieri: Hmm yeah, was just playing with the AES intrinsics in godbolt and the ASM it's generating with a function wrapper isn't quite as good as what invoking the intrinsic does, instead there are some ancillary `ldp`/`stp`/`mov` instructions.

Haven't tried seeing what it does in context with `#[inline(always)]` though... it may be able to elide them in that case?

sha2/src/sha512/aarch64.rs
codahale: It doesn’t seem like it.

Before:

```
test sha512_10    ... bench:           6 ns/iter (+/- 0) = 1666 MB/s
test sha512_100   ... bench:          65 ns/iter (+/- 0) = 1538 MB/s
test sha512_1000  ... bench:         637 ns/iter (+/- 11) = 1569 MB/s
test sha512_10000 ... bench:       6,308 ns/iter (+/- 80) = 1585 MB/s
```

After:

```
test sha512_10    ... bench:           6 ns/iter (+/- 0) = 1666 MB/s
test sha512_100   ... bench:          64 ns/iter (+/- 3) = 1562 MB/s
test sha512_1000  ... bench:         639 ns/iter (+/- 58) = 1564 MB/s
test sha512_10000 ... bench:       6,374 ns/iter (+/- 765) = 1568 MB/s
```

sha2/src/sha256/aarch64.rs
codahale: I wasn’t able to get rid of the copying with `#[inline(always)]` in my AEGIS-128L impl, so I don’t think it’d work here.

tarcieri: Wow, you weren't kidding. I just got `aes` working with polyfill functions and it's about 30% slower versus the intrinsics, even with `#[inline(always)]`

newpavlov: Hm, apart from differences in register allocation, I get identical assembly.

Macros: https://rust.godbolt.org/z/GKfvf4qof

Functions: https://rust.godbolt.org/z/4Gs9Yo56T

sha2/src/sha256/aarch64.rs
```diff
@@ -1,15 +1,138 @@
 //! SHA-256 `aarch64` backend.
 
+// Implementation adapted from mbedtls.
+
 // TODO: stdarch intrinsics: RustCrypto/hashes#257
 
+use core::arch::aarch64::*;
+use core::arch::asm;
```
newpavlov: Small nit, you can write it as:
```rust
use core::arch::{aarch64::*, asm};
```

sha2/src/sha256/aarch64.rs
tarcieri: Interesting! Maybe I'm doing something else wrong then...

codahale: I just tried it and in this context there doesn’t seem to be a penalty, whereas for AES it’s stiff. I wonder if that’s due to the `AESE`/`AESMC` fusion being interrupted?

I’ll change these macros to functions.

sha2/src/sha256/aarch64.rs
```diff
@@ -1,15 +1,138 @@
...
+        unsafe { sha256_compress(state, blocks) }
     } else {
         super::soft::compress(state, blocks);
     }
 }
+
+macro_rules! vsha256hq_u32 {
+    ($hash_efgh:expr, $hash_abcd:expr, $wk:expr) => {{
+        let mut out = $hash_efgh;
+        asm!("SHA256H {:q}, {:q}, {:v}.4S", inout(vreg) out, in(vreg) $hash_abcd, in(vreg) $wk);
```
newpavlov: It may be worth to annotate the assembly snippets with `pure`, `nomem`, `preserves_flags`. I am not sure if `nostack` is needed when `nomem` is already provided.

sha2/src/sha256/aarch64.rs
tarcieri: @codahale that could definitely be it, yeah. Might just need to have a block of inline assembly in the core encryption/decryption functions with both instructions for now

codahale: Ok, they’re now just functions and the performance is exactly the same.

sha2/src/sha256/aarch64.rs
newpavlov: Can you compare performance after addition of the options? The resulting assembly is somewhat different.

newpavlov:
Can you compare performance after addition of the options? The resulting assembly is somewhat different. Number of instructions is the same, so hopefully it's only reordering which may even improve performance.

codahale:
No real difference.

Before:

```
test sha256_10    ... bench:           4 ns/iter (+/- 0) = 2500 MB/s
test sha256_100   ... bench:          46 ns/iter (+/- 2) = 2173 MB/s
test sha256_1000  ... bench:         421 ns/iter (+/- 5) = 2375 MB/s
test sha256_10000 ... bench:       4,155 ns/iter (+/- 44) = 2406 MB/s
test sha512_10    ... bench:           6 ns/iter (+/- 0) = 1666 MB/s
test sha512_100   ... bench:          65 ns/iter (+/- 0) = 1538 MB/s
test sha512_1000  ... bench:         636 ns/iter (+/- 15) = 1572 MB/s
test sha512_10000 ... bench:       6,317 ns/iter (+/- 54) = 1583 MB/s
```

After:
```
test sha256_10    ... bench:           4 ns/iter (+/- 0) = 2500 MB/s
test sha256_100   ... bench:          46 ns/iter (+/- 4) = 2173 MB/s
test sha256_1000  ... bench:         423 ns/iter (+/- 10) = 2364 MB/s
test sha256_10000 ... bench:       4,179 ns/iter (+/- 63) = 2392 MB/s
test sha512_10    ... bench:           6 ns/iter (+/- 0) = 1666 MB/s
test sha512_100   ... bench:          65 ns/iter (+/- 0) = 1538 MB/s
test sha512_1000  ... bench:         636 ns/iter (+/- 12) = 1572 MB/s
test sha512_10000 ... bench:       6,324 ns/iter (+/- 400) = 1581 MB/s
```

newpavlov (approved):

newpavlov:
Thank you!

codahale:
Now that this is merged, what’s the remaining work to drop the `sha2_asm` dependency?

newpavlov:
We would need to migrate the x86-64 implementation from it to inline asm. IIRC it's still a bit faster than our software fallback.

codahale:
I ask b/c I was running some related benchmarks on a GCE `n2-standard-4` with Ice Lake and noticed there wasn’t much of a difference:

With `asm`:

```
$ RUSTFLAGS="-C target-cpu=native" cargo +nightly bench --features=asm

test sha256_10    ... bench:           9 ns/iter (+/- 0) = 1111 MB/s
test sha256_100   ... bench:          81 ns/iter (+/- 0) = 1234 MB/s
test sha256_1000  ... bench:         719 ns/iter (+/- 5) = 1390 MB/s
test sha256_10000 ... bench:       7,126 ns/iter (+/- 64) = 1403 MB/s
test sha512_10    ... bench:          21 ns/iter (+/- 0) = 476 MB/s
test sha512_100   ... bench:         201 ns/iter (+/- 1) = 497 MB/s
test sha512_1000  ... bench:       1,836 ns/iter (+/- 8) = 544 MB/s
test sha512_10000 ... bench:      17,769 ns/iter (+/- 61) = 562 MB/s
```

Without `asm`:

```
$ RUSTFLAGS="-C target-cpu=native" cargo +nightly bench
test sha256_10    ... bench:           9 ns/iter (+/- 0) = 1111 MB/s
test sha256_100   ... bench:          81 ns/iter (+/- 1) = 1234 MB/s
test sha256_1000  ... bench:         718 ns/iter (+/- 4) = 1392 MB/s
test sha256_10000 ... bench:       7,116 ns/iter (+/- 20) = 1405 MB/s
test sha512_10    ... bench:          21 ns/iter (+/- 0) = 476 MB/s
test sha512_100   ... bench:         201 ns/iter (+/- 1) = 497 MB/s
test sha512_1000  ... bench:       1,836 ns/iter (+/- 8) = 544 MB/s
test sha512_10000 ... bench:      17,819 ns/iter (+/- 100) = 561 MB/s
```

Definitely within the margin of error.

Maybe on a different CPU?

newpavlov:
You are getting results for the SHA-NI and AVX2 backends (the asm backend is treated as a replacement for the software backend, thus it has lower priority). On my laptop after I disabled them I get:
```
// without asm:
test sha256_10    ... bench:          33 ns/iter (+/- 0) = 303 MB/s
test sha256_100   ... bench:         321 ns/iter (+/- 2) = 311 MB/s
test sha256_1000  ... bench:       3,131 ns/iter (+/- 6) = 319 MB/s
test sha256_10000 ... bench:      31,227 ns/iter (+/- 69) = 320 MB/s
test sha512_10    ... bench:          21 ns/iter (+/- 0) = 476 MB/s
test sha512_100   ... bench:         207 ns/iter (+/- 0) = 483 MB/s
test sha512_1000  ... bench:       2,017 ns/iter (+/- 5) = 495 MB/s
test sha512_10000 ... bench:      20,077 ns/iter (+/- 64) = 498 MB/s

// with asm:
test sha256_10    ... bench:          28 ns/iter (+/- 3) = 357 MB/s
test sha256_100   ... bench:         274 ns/iter (+/- 7) = 364 MB/s
test sha256_1000  ... bench:       2,671 ns/iter (+/- 23) = 374 MB/s
test sha256_10000 ... bench:      26,693 ns/iter (+/- 348) = 374 MB/s
test sha512_10    ... bench:          20 ns/iter (+/- 0) = 500 MB/s
test sha512_100   ... bench:         184 ns/iter (+/- 2) = 543 MB/s
test sha512_1000  ... bench:       1,809 ns/iter (+/- 14) = 552 MB/s
test sha512_10000 ... bench:      18,032 ns/iter (+/- 177) = 554 MB/s

```

codahale:
Ah, gotcha. Thanks for clearing that up.

sha2/src/sha256/aarch64.rs
tarcieri: @codahale I think your hypothesis about instruction fusions was right. I added a function like this:

```rust
#[inline(always)]
unsafe fn vaeseq_u8_and_vaesmcq_u8(mut data: uint8x16_t, key: uint8x16_t) -> uint8x16_t {
    asm!(
        "AESE {d:v}.16B, {k:v}.16B",
        "AESMC {d:v}.16B, {d:v}.16B",
        d = inout(vreg) data,
        k = in(vreg) key,
        options(pure, nomem, nostack, preserves_flags)
    );
    data
}
```

...and now the performance is the same as before

PR #493, opened 2023-06-27
[WIP] MD6 implementation

iwiwsb:


tarcieri:
Did you close this by mistake? It would still be good to have.

iwiwsb:
@tarcieri My commits were discarded as I pressed 'Sync fork' button. I'm still working on MD6 but I've only implemented the compression function so far. I will open new pull request later when it is ready.

PR #499, opened 2023-08-16
k12: fix incorrect hash calculation

danielreisinger:
If the length of the merged input string is a multiple of the chunk size,
then the previous implementation appends a chaining value for an empty
chunk.


tarcieri:
Thank you!

PR #504, opened 2023-09-20
Add asm!-based backend for LoongArch64 targets

newpavlov:
Based on code from https://github.com/RustCrypto/asm-hashes/pull/66

@heiher 
Can you check whether I translated assembly correctly? Both by running tests and by manual inspection of the code.

Also it would be nice if cross had support for LoongArch64, so we could run tests properly in our CI.

### DIFF
--- a/sha1/src/compress/loongarch64_asm.rs
+++ b/sha1/src/compress/loongarch64_asm.rs
@@ -215,17 +215,17 @@ pub fn compress(state: &mut [u32; 5], blocks: &[[u8; 64]]) {
             "add.w   $t3, $t3, $t8",
             "add.w   $t4, $t4, $t5",
 
-            // Looping over blocks
-            "addi.d  $a1, $a1, 64",
-            "addi.d  $a2, $a2, -1",
-            "bnez    $a2, 42b",
-
             // Save updated state
             "st.w    $t0, $a0, 0",
             "st.w    $t1, $a0, 4",
             "st.w    $t2, $a0, 8",
             "st.w    $t3, $a0, 12",
             "st.w    $t4, $a0, 16",
+
+            // Looping over blocks
+            "addi.d  $a1, $a1, 64",
+            "addi.d  $a2, $a2, -1",
+            "bnez    $a2, 42b",
 
             // Restore stack register
             "addi.d  $sp, $sp, 64",
--- a/sha1/src/compress/loongarch64_asm.rs
+++ b/sha1/src/compress/loongarch64_asm.rs
@@ -30,7 +30,6 @@ macro_rules! scheldule {
     ($i:literal, $e:literal) => {
         c!(
             "ld.w    $t5, $sp, (((" $i " - 3) & 0xF) * 4);"
-            "ld.w    $t5, $sp, (((" $i " - 3) & 0xF) * 4);"
             "ld.w    $t6, $sp, (((" $i " - 8) & 0xF) * 4);"
             "ld.w    $t7, $sp, (((" $i " - 14) & 0xF) * 4);"
             "ld.w    $t8, $sp, (((" $i " - 16) & 0xF) * 4);"
@@ -94,11 +93,11 @@ macro_rules! round3 {
 macro_rules! roundtail {
     ($a:literal, $b:literal, $e:literal, $i:literal, $k:literal) => {
         c!(
-            "ld.w    $t5, $a1, (" $i " * 4);"
-            "revb.2h $t5, $t5;"
-            "rotri.w $t5, $t5, 16;"
+            "rotri.w " $b "," $b ", 2;"
             "add.w " $e "," $e ", $t5;"
-            "st.w    $t5, $sp, (" $i " * 4);"
+            "add.w " $e "," $e "," $k ";"
+            "rotri.w $t5," $a ", 27;"
+            "add.w " $e "," $e ", $t5;"
         )
     };
 }
--- a/sha1/src/compress/loongarch64_asm.rs
+++ b/sha1/src/compress/loongarch64_asm.rs
@@ -109,9 +109,6 @@ pub fn compress(state: &mut [u32; 5], blocks: &[[u8; 64]]) {
     }
 
     unsafe {
-        let mut blocks_ptr = blocks.as_ptr();
-        let mut blocks_len = blocks.len();
-
         asm!(
             // Allocate scratch stack space
             "addi.d  $sp, $sp, -64;",
@@ -234,8 +231,8 @@ pub fn compress(state: &mut [u32; 5], blocks: &[[u8; 64]]) {
             "addi.d  $sp, $sp, 64",
 
             in("$a0") state,
-            inout("$a1") blocks_ptr,
-            inout("$a2") blocks_len,
+            inout("$a1") blocks.as_ptr() => _,
+            inout("$a2") blocks.len() => _,
 
             in("$a4") K[0],
             in("$a5") K[1],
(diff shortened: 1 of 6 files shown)


sha1/src/compress/loongarch64_asm.rs
```diff
@@ -0,0 +1,262 @@
...
+            "addi.d  $sp, $sp, 64",
+
+            in("$a0") state,
+            inout("$a1") blocks_ptr,
+            inout("$a2") blocks_len,
+
+            in("$a4") K[0],
+            in("$a5") K[1],
+            in("$a6") K[2],
+            in("$a7") K[3],
```
newpavlov, line 239: This causes a somewhat different code generation compared to the asm-hashes assembly:
```asm
lu12i.w $a3, -462405
ori     $a6, $a3, 3292
lu12i.w $a3, 370727
ori     $a4, $a3, 2457
lu12i.w $a3, 454046
ori     $a5, $a3, 2977
lu32i.d $a6, 0
lu12i.w $a3, -219604
ori     $a7, $a3, 470
lu32i.d $a7, 0
```

I guess, it's fine?

Alternatively, we could use:
```rust
asm!(
    // ...
    "la.pcrel $t0, {}",
    "ld.w     $a4, $t0, 0",
    "ld.w     $a5, $t0, 4",
    "ld.w     $a6, $t0, 8",
    "ld.w     $a7, $t0, 12",
    // ...
    sym K,
    // ...
);

```

It gets compiled into:
```asm
pcalau12i       $t0, %pc_hi20(example::K)
addi.d  $t0, $t0, %pc_lo12(example::K)
ld.w    $a4, $t0, 0
ld.w    $a5, $t0, 4
ld.w    $a6, $t0, 8
ld.w    $a7, $t0, 12
```

I think the first variant based on immediate constants should by a tiny bit faster, despite twice the number of instructions?

sha1/src/compress/loongarch64_asm.rs
```diff
@@ -0,0 +1,262 @@
...
+            "st.w    $t2, $a0, 8",
+            "st.w    $t3, $a0, 12",
+            "st.w    $t4, $a0, 16",
+
+            // Restore stack register
+            "addi.d  $sp, $sp, 64",
+
+            in("$a0") state,
+            inout("$a1") blocks_ptr,
+            inout("$a2") blocks_len,
```
newpavlov: IIUC it would be incorrect to use:
```rust
in("$a1") blocks.as_ptr(),
in("$a2") blocks.len(),
```

According to the [reference](https://doc.rust-lang.org/reference/inline-assembly.html):

>The allocated register must contain the same value at the end of the asm code (except if a lateout is allocated to the same register).

And the register values get changed in the assembly code. Unfortunately, using `inout` means that we have to silence the unused warning later.

newpavlov: Fixed by using `inout(<reg>) <in expr> => _`.

sha1/src/compress/loongarch64_asm.rs
heiher: I agree with you.

sha1/src/compress/loongarch64_asm.rs
```diff
@@ -0,0 +1,256 @@
...
+            "xor     $t5, $t5, " $d ";"
+            roundtail!($a, $b, $e, $i, "$a4")
+        )
+    };
+}
+
+macro_rules! scheldule {
+    ($i:literal, $e:literal) => {
+        c!(
+            "ld.w    $t5, $sp, (((" $i " - 3) & 0xF) * 4);"
```
heiher, line 32: Remove this redundant line.

sha1/src/compress/loongarch64_asm.rs
```diff
@@ -0,0 +1,256 @@
...
+    ($a:literal, $b:literal, $e:literal, $i:literal, $k:literal) => {
+        c!(
+            "ld.w    $t5, $a1, (" $i " * 4);"
+            "revb.2h $t5, $t5;"
+            "rotri.w $t5, $t5, 16;"
+            "add.w " $e "," $e ", $t5;"
+            "st.w    $t5, $sp, (" $i " * 4);"
+        )
+    };
+}
```
heiher, line 103:
```rust
macro_rules! roundtail {
    ($a:literal, $b:literal, $e:literal, $i:literal, $k:literal) => {
        c!(
            "rotri.w " $b "," $b ", 2;"
            "add.w " $e "," $e ", $t5;"
            "add.w " $e "," $e "," $k ";"
            "rotri.w $t5," $a ", 27;"
            "add.w " $e "," $e ", $t5;"
        )
    };
}
```

heiher (changes requested):
Thanks for your patch.

heiher:
And now

On Loongson 3A5000 @ 2.3GHz:

Without this patch:

```
test sha1_10    ... bench:          33 ns/iter (+/- 0) = 303 MB/s
test sha1_100   ... bench:         308 ns/iter (+/- 0) = 324 MB/s
test sha1_1000  ... bench:       2,559 ns/iter (+/- 2) = 390 MB/s
test sha1_10000 ... bench:      25,057 ns/iter (+/- 11) = 399 MB/s
```

With asm-hashes patch:

```
test sha1_10    ... bench:          32 ns/iter (+/- 0) = 312 MB/s
test sha1_100   ... bench:         291 ns/iter (+/- 0) = 343 MB/s
test sha1_1000  ... bench:       2,450 ns/iter (+/- 1) = 408 MB/s
test sha1_10000 ... bench:      24,079 ns/iter (+/- 12) = 415 MB/s
```

With this patch: :+1: 

```
test sha1_10    ... bench:          31 ns/iter (+/- 0) = 322 MB/s
test sha1_100   ... bench:         286 ns/iter (+/- 0) = 349 MB/s
test sha1_1000  ... bench:       2,408 ns/iter (+/- 1) = 415 MB/s
test sha1_10000 ... bench:      23,637 ns/iter (+/- 12) = 423 MB/s
```

newpavlov:
@heiher 
Your findings are fixed. Can you verify that this implementation passes the tests (do not forget to enable the `loongarch64_asm` feature)? If it does, then I think we can merge and release it.

heiher:
@newpavlov Thank you. The tests are PASSED.

```
$ cargo test -F loongarch64_asm
   Compiling version_check v0.9.4
   Compiling typenum v1.16.0
   Compiling proc-macro-hack v0.5.20+deprecated
   Compiling blobby v0.3.1
   Compiling cfg-if v1.0.0
   Compiling generic-array v0.14.7
   Compiling hex-literal-impl v0.2.3
   Compiling hex-literal v0.2.2
   Compiling crypto-common v0.1.6
   Compiling block-buffer v0.10.4
   Compiling digest v0.10.7
   Compiling sha1 v0.10.5 (/home/hev/rust/hashes/sha1)
    Finished test [optimized + debuginfo] target(s) in 5.48s
     Running unittests src/lib.rs (/home/hev/rust/hashes/target/debug/deps/sha1-efff1781e455b298)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

     Running tests/mod.rs (/home/hev/rust/hashes/target/debug/deps/mod-bf2a6b498c858bf9)

running 2 tests
test sha1_main ... ok
test sha1_rand ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s

   Doc-tests sha1

running 1 test
test sha1/src/lib.rs - (line 14) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.34s
```

PR #510, opened 2023-10-09
blake2: Refuse empty keys in keyed hash construction

### LINKED ISSUE #509: blake2: MAC variants should not be constructable with an empty key?
`blake 0.10` permits construction of MAC variants with empty keys, but I don't think this is a valid construction of blake2. Using [RFC7693](https://www.rfc-editor.org/rfc/rfc7693.html) as reference:

[Section 2.5](https://www.rfc-editor.org/rfc/rfc7693.html#section-2.5) states the following:

> The second (little-endian) byte of the parameter block, "kk", specifies the key size in bytes. Set kk = 00 for unkeyed hashing.

I've to understand this as keyed Blake2 is mutually exclusive with an empty key. Yet `Blake2*Mac` variants seem to permit this construction with [`new_with_params`](https://github.com/RustCrypto/hashes/blob/59176e068a1b79058bfaa6d24a90d2f518b3c1ab/blake2/src/macros.rs#L283). This method will accept an empty key, perform operations to initialize blake2 in keyed mode (e.g. the operations in [Section 3.3](https://www.rfc-editor.org/rfc/rfc7693.html#section-3.3) that assume `kk > 0`), but then [pass a key length of 0 to `new_with_params`](https://github.com/RustCrypto/hashes/blob/59176e068a1b79058bfaa6d24a90d2f518b3c1ab/blake2/src/macros.rs#L297), which then [initializes the parameter block with `kk = 0x00`](https://github.com/RustCrypto/hashes/blob/59176e068a1b79058bfaa6d24a90d2f518b3c1ab/blake2/src/macros.rs#L43).

This results in a keyed construction of blake2 buffers with unkeyed initialization which seems incorrect?

If I'm correct, then I think we either need to special-case the empty key when provided to the MAC variants or have empty keys return an error?

edward-shen:
Fixes #509.

Blake2 0.10 splits keyed hashing into `*Mac` struct variants. However, this permits empty keys to be provided to `new_with_salt_and_personal`, while the implementation assumes a non-empty key. This leads to an invalid construction of Blake2, as the `kk` argument of the parameter block will now be `0x00`, but buffer is initialized as if `kk > 0x00`.

This change introduces a guard to the function, failing if the key was empty.

### DIFF
--- a/blake2/src/macros.rs
+++ b/blake2/src/macros.rs
@@ -276,9 +276,12 @@ macro_rules! blake2_mac_impl {
         {
             /// Create new instance using provided key, salt, and persona.
             ///
-            /// Key length should not be bigger than block size, salt and persona
-            /// length should not be bigger than quarter of block size. If any
-            /// of those conditions is false the method will return an error.
+            /// # Errors
+            ///
+            /// Key length should not be empty or bigger than the block size and
+            /// the salt and persona length should not be bigger than quarter of
+            /// block size. If any of those conditions is false the method will
+            /// return an error.
             #[inline]
             pub fn new_with_salt_and_personal(
                 key: &[u8],
@@ -288,7 +291,7 @@ macro_rules! blake2_mac_impl {
                 let kl = key.len();
                 let bs = <$hash as BlockSizeUser>::BlockSize::USIZE;
                 let qbs = bs / 4;
-                if kl > bs || salt.len() > qbs || persona.len() > qbs {
+                if kl == 0 || kl > bs || salt.len() > qbs || persona.len() > qbs {
                     return Err(InvalidLength);
                 }
                 let mut padded_key = Block::<$hash>::default();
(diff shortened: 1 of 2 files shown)


edward-shen:
Hey @newpavlov! Sorry for the ping; I know your time is limited but I just wanted to check if this is on your or someone else's radar.

I believe this is a pretty easy-to-trigger API hazard (especially for non-experts) so I wanted to make sure this will be addressed sooner than later. Let me know if it's not the case.

Thanks!

tarcieri:
I’d personally like to get #228 over the finish line

edward-shen:
Hey @tarcieri, #228 still requires more discussion right? While that also does resolve #509 from what I can tell, the pace of discussion there feels like it won't be merged soon. Of course, I also might not be privy to discussion elsewhere.

This is a very small change that shouldn't create merge conflicts with #228, since as far as I can tell git is treating your blake2 implementation as completely new files.

I'm pushing for this pretty hard because I feel like preventing incorrect construction of hashes is pretty important, but if the project isn't in the state where it can accept PRs or is not accepting new PRs until #228 is merged, please let me know so I don't annoy you or other maintainers!

newpavlov:
I will merge this PR, but we probably will replace implementation based on #228 before publishing v0.11.

PR #512, opened 2023-10-20
Update examples to use `default()` instead of `new()` which no longer exists

boxbeam:
Also worth noting: `finalize()` does not appear to exist

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Now you can write the following code:
 ```rust
 use sha2::{Sha256, Digest};
 
-let mut hasher = Sha256::new();
+let mut hasher = Sha256::default();
 let data = b"Hello world!";
 hasher.update(data);
 // `update` can be called repeatedly and is generic over `AsRef<[u8]>`
@@ -120,7 +120,7 @@ Instead of calling `update`, you also can use a chained approach:
 ```rust
 use sha2::{Sha256, Digest};
 
-let hash = Sha256::new()
+let hash = Sha256::default()
     .chain_update(b"Hello world!")
     .chain_update("String data")
     .finalize();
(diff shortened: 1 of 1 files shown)


newpavlov:
The methods are part of the [`Digest`](https://docs.rs/digest/latest/digest/trait.Digest.html) trait, which is implemented through the [blanket impl](https://docs.rs/digest/latest/src/digest/digest.rs.html#57-129) for most of hashes in this repository. Also, the doctests are tested as part of our CI, so we know they are valid.

boxbeam:
Might be good to mention that in the examples. I couldn't figure out why new() wasn't working and LSP didn't recommend the trait.

tarcieri:
Which example are you looking at?

The one here says: https://github.com/RustCrypto/hashes#usage

> [`sha2`](https://docs.rs/sha2) and the other hash implementation crates re-export the [`digest`] crate and the [`Digest`] trait for convenience, so you don't have to include it in your `Cargo.toml` it as an explicit dependency.
>
> Now you can write the following code:
>
> ```rust
> use sha2::{Sha256, Digest};
> [...]
> ```
> [`digest`]: https://docs.rs/digest
> [`Digest`]: https://docs.rs/digest/0.10.0/digest/trait.Digest.html

boxbeam:
I guess to me it was unclear that `new` comes from `Digest`.

aetstapa:
Maybe other trait imported caused your problem, if so, the code below should be helpful.
```
use sha2::{Digest, Sha256};
let mut hasher: Sha256 = Digest::new();
Digest::update(&mut hasher, b"hello");
let h = Digest::finalize(hasher);
```
[https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#fully-qualified-syntax-for-disambiguation-calling-methods-with-the-same-name](url)

Issue #87, opened 2019-07-22
Zero memory on drop


fxha:
I have a question about the hash memory management but also for other related repositories. As it seems `RustCrypto` use `GenericArray` and/or raw arrays for sensitive data like keys and internal buffers but the memory is not overwritten/zeroed on reset or when the buffer/struct is dropped for most crates, right? Are there any plans to change this?

newpavlov:
Yes, you are right, currently most crates do not zeroize inner state on drop. IIUC to implement it without hacks we will need either `generic-array` support (e.g. by implementing [`Zeroize`](https://docs.rs/zeroize/0.9.2/zeroize/trait.Zeroize.html) trait.) or const generics.

tarcieri:
@newpavlov you can do a `Zeroize` impl without explicit support for `generic-array` by borrowing the contents of the latter as a byte slice and calling `zeroize()` on that. `IterMut` will also work.

newpavlov:
@tarcieri Maybe we can transform the whole hash state into byte slice (using `unsafe` code) and call zeroize on it?

tarcieri:
@newpavlov or you can just call zeroize on each value in its state in a `Drop` handler. `zeroize` is designed to support structured data. If they're `GenericArray` you just need to call `as_ref()` (or `as_slice()`?) first iirc

When const generics land it should be possible to derive a `Zeroize` impl and drop handler using `zeroize_derive` (the same thing would be possible with first-class `generic-array` support for zeroize, but that doesn't exist).

tarcieri:
@fxha is there a particular hash function or set of them you'd specifically like to have this in?

We can add optional `zeroize` features to high-priority crates.

fxha:
I think it's good practice to (optionally) erase secret data and don't leak keys in memory.

tarcieri:
I agree (in fact I [wrote the most popular crate for it](https://crates.io/crates/zeroize)).

However hash functions are a bit of an odd fish in this regard. Ones based on the Merkle–Damgård construction (which constitute a large number of the ones in this repo, including the popular `sha2`) effectively output their internal state upon finalization (which is what enables length extension attacks). If that internal state were reversible to a secret, that would be a preimage attack against the underlying hash function.

I think the hash functions where this would be relevant are, as you pointed out, ones with keys, i.e. ones which provide PRFs.

As far as I know the only one in this particular repo that applies to is `blake2`.

burdges:
[...]

PR #516, opened 2023-11-12
Implement ZeroizeOnDrop for SHA 1..=2 and Blake2

kayabaNerve:
Relevant to #87.

Resets the hasher state after calling zeroize so it's still usable (1 and 2) indistinguishable from a newly constructed hasher.

This needs a impl Zeroize where Wrapped: Zeroize on CoreWrapper and associated in digest to be pleasantly effective. It's technically reachable now via wrappers' `decompose` functions. I would've done that with this PR yet the digest crate is in a distinct repo.

### DIFF
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["cryptography", "no-std"]
 
 [dependencies]
 digest = { version = "0.10.7", features = ["mac"] }
+zeroize = { version = "1", features = ["derive"], optional = true }
 
 [dev-dependencies]
 digest = { version = "0.10.7", features = ["dev"] }
@@ -20,9 +21,10 @@ hex-literal = "0.2.2"
 
 [features]
 default = ["std"]
-std = ["digest/std"]
+std = ["digest/std", "zeroize?/std"]
 reset = [] # Enable reset functionality
 simd = []
 simd_opt = ["simd"]
 simd_asm = ["simd_opt"]
 size_opt = [] # Optimize for code size. Removes some `inline(always)`
+zeroize = ["dep:zeroize"] # Implement Zeroize for Digest implementors
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -27,4 +27,4 @@ simd = []
 simd_opt = ["simd"]
 simd_asm = ["simd_opt"]
 size_opt = [] # Optimize for code size. Removes some `inline(always)`
-zeroize = ["zeroize_crate"] # Implement Zeroize for Digest implementors
+zeroize = ["zeroize_crate"] # Implement ZeroizeOnDrop for Digest implementors
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -21,7 +21,7 @@ hex-literal = "0.2.2"
 
 [features]
 default = ["std"]
-std = ["digest/std", "zeroize_crate?/std"]
+std = ["digest/std"]
 reset = [] # Enable reset functionality
 simd = []
 simd_opt = ["simd"]
--- a/blake2/Cargo.toml
+++ b/blake2/Cargo.toml
@@ -13,7 +13,7 @@ categories = ["cryptography", "no-std"]
 
 [dependencies]
 digest = { version = "0.10.7", features = ["mac"] }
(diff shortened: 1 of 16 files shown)


newpavlov:
Do not implement the `Zeroize` trait for cryptographic types. Users may execute it and get an invalid state. `Drop` + `ZeroizeOnDrop` will be sufficient.

newpavlov:
Also it's probably worth to wait for the next breaking release of `digest` which will bump MSRV to 1.65 and add `zeroize` support during migration to it.

kayabaNerve:
This does not return an invalid state. It's effectively a reset with a guarantee the prior state is gone. Please read the implementation.

Also, I am unable to get CI happy 0_o

newpavlov:
You literally set zeros into hasher states, which may cause security issues if such state is to be accidentally used by users. For proper reset you should use the initialization constants. You SHOULD NOT implement the `Zeroize` trait for types which may contain and rely on any kind of inner invariant. It's misuse of the trait.

kayabaNerve:
... and then immediately after setting zeroes, set the initialization constants.

kayabaNerve:
If you'd rather it directly set the initialization constants, that can be done. It'd just require re-implementing the zeroize's crate internal mechanisms to avoid being optimized out by the compiler. That would duplicate code (several times), notably increase maintenance effort, and be at greater risk of failure IMO.

newpavlov:
Yes, you are right, but it's not "zeroization" anymore, isn't it? The point of the `Zeroize` trait is to perform "volatile" erasure of secrets and it's usually used to erase "primitive" values. Erasure of "complex" types is done using `Drop` impls which use `Zeroize` under the hood.

I will repeat myself: you should not implement the `Zeroize` trait for the hasher types. See the other algorithm crates which support `zeroize`, e.g. the block cipher crates.

kayabaNerve:
It's not zeroization, yet the intent is to erase the prior state. This does perform exactly that erasure. I'm unsure of anyone who performs zeroization and then reads the memory, always expecting zeroes (no matter the type) like it's a `calloc` call.

As for whether or not this should *solely* implement `ZeroizeOnDrop`... Sure. I'm completely fine with that and can edit this PR to do it that way, per your preference :+1: I only did it this was initially as it leaves the user with greater choice over when they want to zeroize.

tarcieri:
Generally `Zeroize` is useful for the low-level types that store state like integers and slices, and higher-level constructions benefit from `ZeroizeOnDrop` as there's no need to explicitly call `zeroize` (prevents that from being forgotten) and doing it on drop instead of manually prevents a use-after-zeroize condition.

With `ZeroizeOnDrop`, you can just enable the `zeroize` feature and you get zeroization for free with no additional changes.

newpavlov:
The minimal versions CI failures are not relevant to this PR and should be fixed by https://github.com/RustCrypto/actions/pull/34

kayabaNerve:
Happy to hear. I just pushed a commit which solely adds ZeroizeOnDrop for SHA 1..= 2 and Blake2 :+1:

sha3/Cargo.toml
```diff
@@ -26,7 +26,7 @@ hex-literal = "0.2.2"
 
 [features]
 default = ["std"]
-std = ["digest/std"]
+std = ["digest/std", "zeroize?/std"]
```
newpavlov: Remove this line. It causes MSRV breakage.

blake2/Cargo.toml
```diff
@@ -13,6 +13,7 @@ categories = ["cryptography", "no-std"]
 
 [dependencies]
 digest = { version = "0.10.7", features = ["mac"] }
+zeroize_crate = { package = "zeroize", version = "1", default-features = false, optional = true }
```
newpavlov: I think we can simply use `zeroize = { version = "1", default-features = false, optional = true }`?

kayabaNerve: I'm assuming you mean by removing the zeroize feature, which does seem to not be needed in this use-case. Sorry, I'm unfamiliar with writing tomls with such old support :sweat_smile:

kayabaNerve:
Pushed what I believe to be resolutions :+1: Thanks for the tips :)

kayabaNerve:
Is there anything I can do to push this forward?

newpavlov:
Closing in favor of #545.

PR #540, opened 2024-01-10
Reduce amount of `unsafe`

newpavlov:
Also removes no longer necessary `TryInto` imports.


sha2/src/sha512/x86.rs
```diff
@@ -154,7 +154,7 @@ unsafe fn rounds_0_63_avx2(
 
     for i in 1..5 {
         for j in 0..8 {
-            let t = _mm_loadu_si128(K64.as_ptr().add(k64x4_idx) as *const u64 as *const _);
+            let t = _mm_loadu_si128(K64.as_ptr().add(k64x4_idx).cast());
```
tarcieri, line 157: Oh huh, neat

newpavlov: Yeah, it's one of relatively old (introduced in 1.38) goodies which is really nice.

PR #542, opened 2024-01-10
Remove dependency on crates from asm-hashes

newpavlov:
The asm-hashes backends can be a bit faster than the software backends, but they are a maintenance headache and have a bunch of issues like reliance on `cc` and limited cross-platforminess. I think we should eventually archive the asm-hashes repository. Users who desire the additional performance of asm-hashes may either help us to port the code on `asm!`, or stay on the v0.10 crate versions.

MSRV for some crates is bumped to 1.72 because of the Loongarch `asm!` backends, which are now enabled by default.

### DIFF
--- a/.github/workflows/md5.yml
+++ b/.github/workflows/md5.yml
@@ -24,7 +24,7 @@ jobs:
   set-msrv:
     uses: RustCrypto/actions/.github/workflows/set-msrv.yml@master
     with:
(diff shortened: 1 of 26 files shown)


tarcieri (approved):
Nice. Removing the old ASM backends should make it easier to add inline backends as a replacement.

PR #556, opened 2024-01-28
Updates to remove `hybrid_array::ArrayOps`

tarcieri:
This was a breaking change: RustCrypto/hybrid-array#30

This PR includes the necessary updates to remain compatible.


tarcieri:
I've made this a draft so we can bump the `hybrid-array` version after a release

PR #564, opened 2024-02-10
Blake2: add missing types of common digest sizes

HeCorr:
See comment [#67](https://github.com/RustCrypto/hashes/issues/67#issuecomment-1328463458).

This PR adds the following common fixed digest sizes:
- `BLAKE2s-128`
- `BLAKE2b-128`
- `BLAKE2b-256`

Unsure if I should also add the Mac variants, please let me know!

### DIFF
--- a/blake2/src/lib.rs
+++ b/blake2/src/lib.rs
@@ -18,7 +18,7 @@ use core::{fmt, marker::PhantomData, ops::Div};
 use digest::{
     array::{Array, ArraySize},
     block_buffer::{Lazy, LazyBuffer},
-    consts::{U128, U32, U4, U64},
+    consts::{U128, U16, U32, U4, U64},
     core_api::{
         AlgorithmName, Block, BlockSizeUser, Buffer, BufferKindUser, CoreWrapper,
         CtVariableCoreWrapper, OutputSizeUser, RtVariableCoreWrapper, TruncSide, UpdateCore,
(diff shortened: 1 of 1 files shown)


HeCorr:
Also, I've noticed that only the `BLAKE2b-512` fixed variant is [tested](https://github.com/RustCrypto/hashes/blob/c38787b7abd8dd10412972941d077df54c95a4bb/blake2/tests/mod.rs#L7), but I've just recently started learning Rust so I'm not sure how to write tests for the other variants (specifically how to generate the .blb files).

Any guidance on this would be very welcome!

StackOverflowExcept1on (approved):

newpavlov:
Could you please link cases where those variations are used in practice?

>only the BLAKE2b-512 fixed variant is tested

We could add tests for other variants, but we would have to generate tests ourselves. Ideally, we would use "official" test vectors, but we probably couldn't find them at the time.

HeCorr:
> Could you please link cases where those variations are used in practice?

I can't, but I personally always use either 256 or 128 if I need the hash to be shorter. What I can link you though are other projects which do include them and even other less common sizes (160, 384..):

[CyberChef](https://gchq.github.io/CyberChef/#recipe=BLAKE2b('256','Hex',%7B'option':'UTF8','string':''%7D)&input=aW5wdXQ)
[Go standard lib](https://pkg.go.dev/golang.org/x/crypto/blake2b)

StackOverflowExcept1on:
I have similar problem that I need hash of 256 bit size, but this crate does not provide BLAKE2b-256.

HeCorr:
> this crate does not provide BLAKE2b-256.

@StackOverflowExcept1on this is how you can currently generate a Blake2b-256 hash:
```Rust
use blake2::{Blake2b, Digest, digest::consts::U32};

type Blake2b256 = Blake2b<U32>;

let mut hasher = Blake2b256::new();
hasher.update(b"my_input");
let res = hasher.finalize();
```

PR #566, opened 2024-02-22
sha1: implement collision detection and mitigation 

### LINKED ISSUE #204: Adding collision detection
Like it or not, SHA-1 is still used in practice.  In [Sequoia PGP](https://sequoia-pgp.org), we initially decided to bad list SHA-1.  However, a developer of Secure Drop recently approached us and asked us to reconsider this decision.  They have users who use certificates secured by SHA-1.  And, unfortunately, they don't have processes to update these offline certificates.  But, even if they did, they doubt that their users would implement them ("never change a running system").

This motivated me to see who else is using SHA-1 in the OpenPGP ecosystem.  My results are tabulated [here](https://gitlab.com/sequoia-pgp/sequoia/-/issues/595).  The results are depressing.  The short version is: Microsoft's Security Notification certificate (created Oct. 31, 2019!) uses SHA-1. 106 of the 884 certificates in the Debian keyring are completely unusable without accepting SHA-1, and another 80 have at least one User ID or subkey that uses SHA-1! 32 of the 76 Arch packagers' certificates are unusable without accepting SHA-1, and 14 further certificates rely on SHA-1 for at least one User ID or subkey.  I also discovered that Fedora misconfigured the tooling used to generate their release keys: although they generate 4k RSA keys, they rely exclusively on SHA-1!  (The configuration is now fixed.)

Unfortunately, [the OpenPGP ecosystem is not in a good place](https://mailarchive.ietf.org/arch/msg/openpgp/NJpWerkG-V1AMxQ-wGHp1eOSwFc) to get people to fix their certificates.  First, most users don't even know there is a problem. Due to the publication of [SHA-1 is a shambles](https://sha-mbles.github.io/), GnuPG now rejects new *third-party* certifications made using SHA-1, but self signatures and binding signatures using SHA-1 are still accepted without warning.  RNP (used by TB) appears to accept anything.  They [only](https://bugzilla.mozilla.org/show_bug.cgi?id=1641720#c3) [plan](https://github.com/rnpgp/rnp/issues/1281) to add a warning when SHA-1 is used.  Second, updating an existing certificate to not use SHA-1 is hard, and [not always entirely possible using GnuPG](https://dev.gnupg.org/T5110)!

This creates enormous pressure on us (Sequoia) to also accept SHA-1.  Initially, we thought that we could accept SHA-1 for self signatures, because they only need to be preimage resistant.  But, [it appears](https://gitlab.com/sequoia-pgp/sequoia/-/issues/595#note_434030365) that there might be a variant of the impersonation attack presented in the SHA-1 is a Shambles paper that could use a collision to forge a self signature.

A different approach would be to use counter cryptanalysis techniques.  This idea and their application to MD5 and SHA-1 was published by [Marc Stevens](https://www.marc-stevens.nl/research/papers/C13-S.pdf).  He also published a [library](https://github.com/cr-marcstevens/sha1collisiondetection) that implements these mitigations for SHA-1.  According to the README:

> This library and command line tool were designed as near drop-in replacements for common SHA-1 libraries and sha1sum. They will compute the SHA-1 hash of any given file and additionally will detect cryptanalytic collision attacks against SHA-1 present in each file. It is very fast and takes less than twice the amount of time as regular SHA-1.
> 
> More specifically they will detect any cryptanalytic collision attack against SHA-1 using any of the top 32 SHA-1 disturbance vectors with probability 1:
> ...
> The possibility of false positives can be neglected as the probability is smaller than 2^-90.
> 
> The library supports both an indicator flag that applications can check and act on, as well as a special safe-hash mode that returns the real SHA-1 hash when no collision was detected and a different safe hash when a collision was detected.

This library is written in C and licensed under the MIT software license.  It would be great if Rust Crypto could include these techniques.

I see two ways to do this: 1. The hash functions are changed to be fallible. 2. A parallel interface is added.

dignifiedquire:
This implements sha1 collision detection, including rehashing for mitigation.
As the code is 1-1 based on the version that git uses, the mitigation hashes should match as well.

Closes #204 

## Open TODOs

- [x] API
- [x] More code cleanup
- [x] Finish integrating of the test cases
- [x] Happy CI
- [x] Benchmarks

## Limitations

Can only be used with the pure software implementation, asfaiu. The reason for this is, that the algorithm needs access to all  intermediary states, and so processing 4 rounds at once through the intrinsics will screw things up. 
For that reason I have made it it's own implementation, instead of adapting the existing `compress` implementations.

It might be possible to add support for the "simpler" assembly implementations that do round for round processing, but I think this could be a follow up in the future, if this is too slow for these platforms.

## Prior art 

- Paper: https://marc-stevens.nl/research/papers/C13-S.pdf
- C reference implementation: https://github.com/cr-marcstevens/sha1collisiondetection
- C2Rust translation: https://gitlab.com/sequoia-pgp/sha1collisiondetection/


sha1/tests/mod.rs
```diff
@@ -13,3 +13,89 @@ fn sha1_rand() {
...
+#[cfg(all(feature = "collision", not(feature = "force-soft")))]
+#[test]
+fn shattered_2() {
+    collision_test(
+        &include_bytes!("./data/shattered-2.pdf")[..],
+        hex!("38762cf7f55934b34d179ae6a4c80cadccbb7f0a"),
+        hex!("e1761773e6a35916d99f891b77663e6405313587"),
+    )
+}
+
```
Byron: I think an explicit test for the default values of `checked::Config`, somewhere, would have helped me in understanding the `collision_test!` macro, as it wasn't obvious to me that many of these flags that are turned off, are on by default.

dignifiedquire: update the api to use a builder pattern, which should make it clearer

dignifiedquire:
Benchmarks on my macbook (arm M1), using the defaults, so both detection and mitigation enabled.

```
test sha1_10              ... bench:          12 ns/iter (+/- 0) = 833 MB/s
test sha1_100             ... bench:         107 ns/iter (+/- 3) = 934 MB/s
test sha1_1000            ... bench:       1,055 ns/iter (+/- 48) = 947 MB/s
test sha1_10000           ... bench:      10,521 ns/iter (+/- 659) = 950 MB/s
test sha1_collision_10    ... bench:          19 ns/iter (+/- 1) = 526 MB/s
test sha1_collision_100   ... bench:         180 ns/iter (+/- 6) = 555 MB/s
test sha1_collision_1000  ... bench:       1,660 ns/iter (+/- 55) = 602 MB/s
test sha1_collision_10000 ... bench:      16,414 ns/iter (+/- 509) = 609 MB/s
```

sha1/tests/mod.rs
Byron: Unfortunately I am spoiled already, knowing what I know. However, since it's the 'checked' version of the hash, I think it's absolutely fair to assume that all checking features are enabled by default unless they are disabled explicitly. The builder pattern doesn't make the more discoverable in tests I think (but an assertion for the desired value of the fields would have made it clear).

Using a builder for that might be a matter of whether that pattern was used in this crate before or not. I'd personally lean towards fields in a struct mainly because they are well-known, won't change (it's all part of the spec implemented here), and are easy enough to use.

Byron (commented):
> Benchmarks on my macbook (arm M1), using the defaults, so both detection and mitigation enabled.
> 
> ```
> test sha1_10              ... bench:          12 ns/iter (+/- 0) = 833 MB/s
> test sha1_100             ... bench:         107 ns/iter (+/- 3) = 934 MB/s
> test sha1_1000            ... bench:       1,055 ns/iter (+/- 48) = 947 MB/s
> test sha1_10000           ... bench:      10,521 ns/iter (+/- 659) = 950 MB/s
> test sha1_collision_10    ... bench:          19 ns/iter (+/- 1) = 526 MB/s
> test sha1_collision_100   ... bench:         180 ns/iter (+/- 6) = 555 MB/s
> test sha1_collision_1000  ... bench:       1,660 ns/iter (+/- 55) = 602 MB/s
> test sha1_collision_10000 ... bench:      16,414 ns/iter (+/- 509) = 609 MB/s
> ```


Thanks a lot for adding a benchmark, great for a better feel for what to expect.

I think Git gets about 1.2GB/s, and `gitoxide` is in the same ballpark when the ASM version of the implementation is used (which is great). Somehow, Git manages to do that with collision detection (probably), it's quite a bit magical.

However, it's without any doubt great to have the implementation and probably it also has to be the new default as correctness has to win, and those who can or want will get a switch to turn it off to get back to the previous (current) behaviour where collisions aren't checked for at all.

This will probably mean that a typical kernel pack, that decompresses to ~100GB, will need ~80s more CPU time. But probably I should actually test this with the version from this branch as other factors may matter as well, like maintaining the additional state that can start to accumulate if there is more than 1m streams hashed per second (but on 10 cores with one hasher per core).

I shall be back (but that really shouldn't hold up this PR at all).

Byron:
Here is my measurements, for informative purposes only, to demonstrate the pack resolution performance - an operation that happens for every clone or fetch.

##### Baseline

This is comparing the assembly version with the one without assembly, with the default sha1 0.10.6 crate from crates.io.

```
❯ hyperfine -w1 -N "gix free pack verify $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx" "./target/release/gix free pack verify $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx" -r2
Benchmark 1: gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
  Time (mean ± σ):      9.454 s ±  0.061 s    [User: 68.773 s, System: 2.390 s]
  Range (min … max):    9.411 s …  9.497 s    2 runs

Benchmark 2: ./target/release/gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
  Time (mean ± σ):     16.110 s ±  0.136 s    [User: 132.938 s, System: 2.582 s]
  Range (min … max):   16.013 s … 16.206 s    2 runs

Summary
  gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx ran
    1.70 ± 0.02 times faster than ./target/release/gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
```

With default features, ASM didn't really kick in apparently.

#### Collision Detection

With the `collision` feature enabled.

```
❯ hyperfine -w1 -N "gix free pack verify $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx" "./target/release/gix free pack verify $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx" -r2
Benchmark 1: gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
  Time (mean ± σ):      9.487 s ±  0.017 s    [User: 68.135 s, System: 2.583 s]
  Range (min … max):    9.475 s …  9.499 s    2 runs

Benchmark 2: ./target/release/gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
  Time (mean ± σ):     26.471 s ±  0.263 s    [User: 225.025 s, System: 3.169 s]
  Range (min … max):   26.285 s … 26.656 s    2 runs

Summary
  gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx ran
    2.79 ± 0.03 times faster than ./target/release/gix free pack verify /Users/byron/dev/github.com/Byron/gitoxide/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
```

#### Comparison by hand

```
gitoxide ( sha1-collisions) +1 -1 [$!?] took 26s
❯ ./target/release/gix free pack verify $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
 20:54:39 Hash of index 'pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx' done 209.7MB in 0.44s (477.7MB/s)
 20:54:39                                           collecting sorted index done 7.5M entries in 0.99s (7.5M entries/s)
 20:54:41 Hash of pack 'pack-b73c0dd365b8ab881a163a1d342dac7416f52591.pack' done 1.3GB in 3.13s (423.2MB/s)
 20:54:41                                                          indexing done 7.5M objects in 2.15s (3.5M objects/s)
 20:55:05                                                         Resolving done 7.5M objects in 23.75s (315.4k objects/s)
 20:55:05                                                          Decoding done 94.3GB in 23.75s (4.0GB/s)

gitoxide ( sha1-collisions) +1 -1 [$!?] took 27s
❯ gix free pack verify $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
 20:55:22 Hash of index 'pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx' done 209.7MB in 0.14s (1.5GB/s)
 20:55:23 Hash of pack 'pack-b73c0dd365b8ab881a163a1d342dac7416f52591.pack' done 1.3GB in 0.78s (1.7GB/s)
 20:55:23                                           collecting sorted index done 7.5M entries in 0.83s (9.1M entries/s)
 20:55:24                                                          indexing done 7.5M objects in 1.36s (5.5M objects/s)
 20:55:32                                                         Resolving done 7.5M objects in 7.37s (1.0M objects/s)
 20:55:32                                                          Decoding done 94.3GB in 7.37s (12.8GB/s)
```

We are seeing a little less than 1/3rd of the ASM performance.

#### Summary 

Here I ran out of time as for some reason, when patching in the modified version of the crate alone (which required a clone and a version modification to match the one available locally), it triggered a lot of compile errors across the codebase that had nothing to do with it. Usually along the line of `.as_ref()` can't be inferred, needs type annotation. I have no explanation for this.

However, if the benchmarks above are an indication, we might be talking 609/950 (roughly 64%) reduction of this particular operation.

I truly wonder how [Git can be this fast](https://github.com/git/git/blob/master/sha1dc/sha1.c#L1747), as its hash performance doesn't seem to need ASM and seemingly supports collision detection as well.


#### Usability in `gitoxide`

It can probably be used with a runtime switch, but I didn't yet test if the increased footprint (due to the check-state) has a negative effect on performance even when it's disabled. It's probably hard to say much about this if ASM doesn't kick in automatically anymore in version 0.11. Technically it will be usable for sure.

dignifiedquire:
> and `gitoxide` is in the same ballpark when the ASM version of the implementation is used

Do you have a link to that ASM code, would love to check out how they do it, after simplifying the code a bit, I think there is probably more that can be done, but adding an optimised version certainly can come later.

dignifiedquire:
> I truly wonder how [Git can be this fast](https://github.com/git/git/blob/master/sha1dc/sha1.c#L1747), as its hash performance doesn't seem to need ASM and seemingly supports collision detection as well.

This is quite surprising indeed, as the implementation here is based on the exact same code

dignifiedquire:
When running on my linux machine with an AMD that has Sha intrinsics 

```
test sha1_10              ... bench:           7 ns/iter (+/- 0) = 1428 MB/s
test sha1_100             ... bench:          58 ns/iter (+/- 0) = 1724 MB/s
test sha1_1000            ... bench:         516 ns/iter (+/- 1) = 1937 MB/s
test sha1_10000           ... bench:       5,091 ns/iter (+/- 24) = 1964 MB/s
test sha1_collision_10    ... bench:          27 ns/iter (+/- 0) = 370 MB/s
test sha1_collision_100   ... bench:         250 ns/iter (+/- 2) = 400 MB/s
test sha1_collision_1000  ... bench:       2,322 ns/iter (+/- 12) = 430 MB/s
test sha1_collision_10000 ... bench:      22,969 ns/iter (+/- 138) = 435 MB/s
```

Byron:
> > and `gitoxide` is in the same ballpark when the ASM version of the implementation is used
> 
> Do you have a link to that ASM code, would love to check out how they do it, after simplifying the code a bit, I think there is probably more that can be done, but adding an optimised version certainly can come later.

These can be activated by adding the `asm` feature to the `sha1` crate, at least in version 0.10. I had a feeling these don't exist like this anymore in 0.11. And I definitely agree, optimization can come later.
At this performance loss, `gitoxide` will probably work hard to get the SHA256 transition going 😅.

dignifiedquire:
> These can be activated by adding the `asm` feature to the `sha1` crate, at least in version 0.10.

Ah yeah, I know more about these than I want to actually 🤣 , on `0.11` it's all auto detected by default.
I thought there might be another hash implementation somewhere 😅 

The regular ASM implementation could be ported to do detection I believe, as it does rounds manually (just checked). But the overall work is still around 1.5x at least, soo... 
The current code could be probably tweaked a bit more to make the rust compiler happier which should help some.

Byron:
> Ah yeah, I know more about these than I want to actually 🤣 , on `0.11` it's all auto detected by default.
> I thought there might be another hash implementation somewhere 😅

There can only be one ;)! But in all seriousness, it's great to hear as it would remove one reason for incompatibility that was plaguing `gitoxide` and required a workaround. I hope it will still be as fast, and will definitely be checking it :)..

From my tests, I have filled them [into the original comment](https://github.com/RustCrypto/hashes/pull/566#issuecomment-1961804338), performance (with collision-checks) looks like close to 1/3rd of what it is with ASM, while the latest version, 0.11 with ASM feature detection, seems to not kick  in ASM on my machine (M1) or with the `--release` build settings I was using. So either way, performance was greatly reduced.

But there is no way around biting that bullet, even though I will definitely put in an escape hatch for those who want to live like it's 2016😅.

In any case, and in case it's not clear over my ramblings about performance: Absolutely stunning work, I am loving it! (*it's just terrible to realize that a lot of the advertised performance came from not comparing apples to apples*)

----

In case it matters, I also *had* to make these unrelated changes because the Rust compiler didn't like it anymore when using the code in this PR. I have no explanation.

```diff
commit 798b11a57977e353ecb53cd7b85f0525e74c15d0
Author: Sebastian Thiel <sebastian.thiel@icloud.com>
Date:   Fri Feb 23 19:00:41 2024 +0100

    TBD

diff --git a/gix-object/src/commit/message/body.rs b/gix-object/src/commit/message/body.rs
index 0da6d1d14..6e84f3e2e 100644
--- a/gix-object/src/commit/message/body.rs
+++ b/gix-object/src/commit/message/body.rs
@@ -33,7 +33,7 @@ pub struct TrailerRef<'a> {
 
 fn parse_single_line_trailer<'a, E: ParserError<&'a [u8]>>(i: &mut &'a [u8]) -> PResult<(&'a BStr, &'a BStr), E> {
     *i = i.trim_end();
-    let (token, value) = separated_pair(take_until(1.., b":".as_ref()), b": ", rest).parse_next(i)?;
+    let (token, value) = separated_pair(take_until(1.., &b":"[..]), b": ", rest).parse_next(i)?;
 
     if token.trim_end().len() != token.len() || value.trim_start().len() != value.len() {
         Err(winnow::error::ErrMode::from_error_kind(i, ErrorKind::Fail).cut())
diff --git a/gix-pack/src/data/input/bytes_to_entries.rs b/gix-pack/src/data/input/bytes_to_entries.rs
index 7450e9134..e089250eb 100644
--- a/gix-pack/src/data/input/bytes_to_entries.rs
+++ b/gix-pack/src/data/input/bytes_to_entries.rs
@@ -138,7 +138,7 @@ where
 
         let crc32 = if self.compressed.crc32() {
             let mut header_buf = [0u8; 12 + gix_hash::Kind::longest().len_in_bytes()];
-            let header_len = entry.header.write_to(bytes_copied, &mut header_buf.as_mut())?;
+            let header_len = entry.header.write_to(bytes_copied, &mut header_buf.as_mut_slice())?;
             let state = gix_features::hash::crc32_update(0, &header_buf[..header_len]);
             Some(gix_features::hash::crc32_update(state, &compressed))
         } else {
diff --git a/gix-pack/src/data/input/entry.rs b/gix-pack/src/data/input/entry.rs
index 7d3d9b3cb..cb5c2af80 100644
--- a/gix-pack/src/data/input/entry.rs
+++ b/gix-pack/src/data/input/entry.rs
@@ -33,7 +33,7 @@ impl input::Entry {
         let mut header_buf = [0u8; 12 + gix_hash::Kind::longest().len_in_bytes()];
         let header_len = self
             .header
-            .write_to(self.decompressed_size, &mut header_buf.as_mut())
+            .write_to(self.decompressed_size, &mut header_buf.as_mut_slice())
             .expect("write to memory will not fail");
         let state = gix_features::hash::crc32_update(0, &header_buf[..header_len]);
         gix_features::hash::crc32_update(state, self.compressed.as_ref().expect("we always set it"))
diff --git a/gix-worktree/src/stack/state/mod.rs b/gix-worktree/src/stack/state/mod.rs
index 30e3c609f..07bffbfbc 100644
--- a/gix-worktree/src/stack/state/mod.rs
+++ b/gix-worktree/src/stack/state/mod.rs
@@ -96,7 +96,7 @@ impl State {
         let a1_backing;
         #[cfg(feature = "attributes")]
         let a2_backing;
-        let names = match self {
+        let names: &[(&bstr::BStr, Option<_>)] = match self {
             State::IgnoreStack(ignore) => {
                 a1_backing = [(
                     ignore.exclude_file_name_for_directories.as_bytes().as_bstr(),
```

dignifiedquire:
Two questions, (1) do you happen to have performance numbers for the version `git` is using? (2) do you know which configuration for the collision detection they are running?

Byron:
Regarding 1): No, and it's not truly comparable as Git has a lot of extra overhead and it doesn't scale with cores at all. Maybe in single-core mode it can be somewhat comparable though, especially with a profiler to see where it spends time.

`git verify-pack $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx` takes 51s (but only uses 3 cores), maybe it does more as well, I don't know.

If that is normalized to a single core with `git -c pack.threads=1 verify-pack $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx` it takes 1m25s, pretty good still, whereas `gix` (with collisions enabled) takes 3m13s (`./target/release/gix -t1 free pack verify $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx`). Interestingly, the current highest performing build can do it (with a single core) in 1m1s.

So I think the 1.5x (better 1.4x) number for the ASM version seems to be what Git has, more or less, assuming all else equal. It's a respectable result that Git presents on a single core.

Regarding 2) All I know is the source in `sha1.c` which I linked earlier and I think you have seen already. There can always be compile-time tricks though that I don't understand, and the local version I am using is `git version 2.39.3 (Apple Git-145)`, whatever magic sauce that comes with.

Just to be sure it wasn't laced with too much fairy dust, I built e02ecfcc534e2021aae29077a958dd11c3897e4c of `git` myself and ran `/Users/byron/dev/github.com/git/git/git -c pack.threads=1 verify-pack $PWD/tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx` in 1m 24s, so it's achieving the same result. Maybe it's just that C code compiling to something very, very fast?

dignifiedquire:
have you tried building with rustflags="target-cpu=native"
this makes quite a big difference in these scenarios

Byron:
That's a valid point! It doesn't make a difference for this branch, it's virtually unchanged. After updating to the most recent version in this branch I did notice a 5% speedup though.
When retrying the latest `master` in this repository, its 820MB/s SHA1 performance still couldn't get close to the 1.7GB I get with the current implementation.

I think what I will end up with is to use v0.10 for the performance-toggle, and the checked version here for everything else. 

Also, I couldn't resist to make measurements on Git's hashing alone, using a 200MB file as input. 

```
❯ hyperfine -N -w1 'git hash-object -t blob tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx'
Benchmark 1: git hash-object -t blob tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.idx
  Time (mean ± σ):     113.6 ms ±   2.3 ms    [User: 92.8 ms, System: 14.5 ms]
  Range (min … max):   112.4 ms … 124.2 ms    26 runs
```

If that is extrapolated, we get to a whopping 1.5GB/s - 1.7GB/s.

With a bigger file of 1326MB, I get:

```
❯ hyperfine -N -w1  'git hash-object --literally tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.pack'
Benchmark 1: git hash-object --literally tests/fixtures/repos/linux.git/objects/pack/pack-b73c0dd365b8ab881a163a1d342dac7416f52591.pack
  Time (mean ± σ):     806.9 ms ±   8.6 ms    [User: 554.0 ms, System: 180.9 ms]
  Range (min … max):   795.6 ms … 820.0 ms    10 runs
```

Which puts it at around 1.64GB/s .

And because I couldn't believe that a bunch of C in a file with a casual build can be this fast, I profiled the run with the 1.3GB file, and here is the answer to all of our questions.

Screenshot 2024-02-24 at 09 42 28

The entrypoint in Git seems to [be here](https://github.com/git/git/blob/3c2a3fdc388747b9eaf4a4a4f2035c1c9ddb26d0/compat/sha1-chunked.c#L4), which clearly talks about platform specific implementations. All the configuration for these seems to [come from here](https://github.com/git/git/blob/master/hash-ll.h#L0-L1).

Maybe they deal with shattered, maybe they don't? I don't know how to validate that as unfortunately, `git hash-object` will add the Git object header to any input it's provided. But since the software fallback for Sha1 contains the mitigation, I presume they wouldn't use a platform-specific acceleration if it wouldn't have the same qualities.

That was an interesting excursion, I learned something. Maybe in the end, in order to be competitive, I'd need a `gix-sha1` crate that uses `sha1` with mitigration as fallback, like Git, but links to the platform variants when possible.

dignifiedquire:
Okay, that hash is not using collision detection, that is using the regular accelerated sha1 routines, which is why it can be so fast. I suspect what git is doing is that it only uses the hash collision version on very specific points, and not internally, to avoid the slow down.

dignifiedquire:
The regression comes from here: https://github.com/RustCrypto/hashes/pull/542 the asm versions are actually fully removed :( 

I guess it really is time to port the assembly to `asm!`..

Byron:
> Okay, that hash is not using collision detection, that is using the regular accelerated sha1 routines, which is why it can be so fast. I suspect what git is doing is that it only uses the hash collision version on very specific points, and not internally, to avoid the slow down.

I looked for 15 minutes and could only conclude that this seems to be a compile-time-only affair. Git doesn't always get compiled with its own implementation, but if enabled it would detect collisions and [die when one was found](https://github.com/git/git/blob/master/sha1dc_git.c#L19). But I might well be wrong with that. It feels a bit like the mitigation was quickly made available, but over time was supplanted by faster implementations again.

Git also doesn't have a choice for this beyond compile time it seems (or I can't find it), which makes me think that if, for some reason, the runtime switching doesn't work or reduces performance, I could also go back to a compile time switch and probably would reduce complexity that way. And I am saying this in case it would simplify things here as well to go compile-time only.

dkg:
This is really great work.  Thank you both for the thoughtful back-and-forth, for documenting your sleuthing here, and for valuing correctness over speed.

sha1/src/checked/ubc_check.rs
```diff
@@ -0,0 +1,1179 @@
+//! Direct translation of the C code
+//! Original license and comments:
+//!
+//! Copyright 2017 Marc Stevens <marc@marc-stevens.nl>, Dan Shumow <danshu@microsoft.com>
+//! Distributed under the MIT Software License.
+//! See accompanying file LICENSE.txt or copy at
+//! https://opensource.org/licenses/MIT
```
tarcieri: This code is MIT only which slightly changes the licensing situation for this crate, which is otherwise Apache 2.0+MIT.

We could note that the code for the `collision` feature is MIT-only in the toplevel README.md, or potentially ask the original authors if they would be okay with the resulting translation being dual licensed as Apache 2.0 in addition to MIT.

sha1/src/lib.rs
```diff
@@ -26,6 +26,12 @@ use digest::const_oid::{AssociatedOid, ObjectIdentifier};
 #[cfg(feature = "zeroize")]
 use digest::zeroize::{Zeroize, ZeroizeOnDrop};
 
+#[cfg(feature = "std")]
+extern crate std;
+
+#[cfg(feature = "collision")]
+pub mod checked;
```
tarcieri: I think it would make sense for this module name to match the feature name, e.g. either call them both `collision`, `checked`, or something else consistent

sha1/src/checked/ubc_check.rs
```diff
@@ -0,0 +1,1179 @@
...
+//!
+//! this file was generated by the 'parse_bitrel' program in the tools section
+//! using the data files from directory 'tools/data/3565'
+//!
+//! sha1_dvs contains a list of SHA-1 Disturbance Vectors (DV) to check
+//! dvType, dvK and dvB define the DV: I(K,B) or II(K,B) (see the paper)
+//! dm[80] is the expanded message block XOR-difference defined by the DV
+//! testt is the step to do the recompression from for collision detection
+//! maski and maskb define the bit to check for each DV in the dvmask returned by ubc_check
```
tarcieri: I think it would be helpful if this were moved to be a doc comment on `SHA1_DVS`

sha1/src/checked/ubc_check.rs
```diff
@@ -0,0 +1,1179 @@
...
+//!
+//! ubc_check takes as input an expanded message block and verifies the unavoidable bitconditions for all listed DVs
+//! it returns a dvmask where each bit belonging to a DV is set if all unavoidable bitconditions for that DV have been met
+//! thus one needs to do the recompression check for each DV that has its bit set
```
tarcieri: Similar to above, this should probably be moved to be a doc comment on `fn ubc_check`

sha1/src/checked/ubc_check.rs
tarcieri: Also is `testt` a typo?

sha1/src/lib.rs
dignifiedquire: yes makes sense, I don't like either name particularly, want to do a tiebreaker?

sha1/src/checked/ubc_check.rs
dignifiedquire: > Also is `testt` a typo?

I don't think so, it is the name used everywhere in the original code

sha1/src/checked/ubc_check.rs
dignifiedquire: @cr-marcstevens according to the license file you are the original license holder, would you be okay with us relicensing under apache 2.0 + MIT?

sha1/src/checked/ubc_check.rs
tarcieri: Maybe put backticks around it so it looks intentional?

sha1/src/lib.rs
tarcieri: `collision-detection` / `collision_detection` would be a lot clearer IMO, albeit verbose

sha1/src/checked/ubc_check.rs
dignifiedquire: also @shumow

sha1/src/checked/ubc_check.rs
dignifiedquire: fixed

sha1/src/checked/ubc_check.rs
dignifiedquire: fixed

sha1/src/lib.rs
```diff
@@ -62,38 +70,77 @@ impl UpdateCore for Sha1Core {
     fn update_blocks(&mut self, blocks: &[Block<Self>]) {
         self.block_len += blocks.len() as u64;
         let blocks = Array::cast_slice_to_core(blocks);
-        compress(&mut self.h, blocks);
+
+        #[cfg(feature = "collision")]
+        {
+            if let Some(ref mut ctx) = self.detection {
```
newpavlov: I am not sure we should modify the "original" SHA-1 implementation in such fashion. We probably should keep it vulnerable and strongly advice to use the hardened version if possible in the crate docs.

dignifiedquire: fixed, it is now fully standalone

sha1/src/checked/ubc_check.rs
cr-marcstevens: Hi Friedel, Tony, 
Thanks for your nice work making our code also available in rust!
And I'm certainly fine with dual licensing Apache 2.0+MIT (with the proper attribution and copyright notices). 
But wouldn't the same situation also apply to the rust translation of our `sha1.c` code?

dignifiedquire: > And I'm certainly fine with dual licensing Apache 2.0+MIT (with the proper attribution and copyright notices).

awesome, thank you!

> But wouldn't the same situation also apply to the rust translation of our sha1.c code?

Yes very much, I will be doing a final pass of notes for attribution on the relevant files in a bit. I'll ping you when it is ready for review.

dignifiedquire: I have added a detailed reference to `src/checked.rs` and pointers to this and the original files in both `checked/ubc_check.rs` and `checked/compress.rs`. Please let me know if this is good for you

sha1/src/checked.rs
```diff
@@ -0,0 +1,386 @@
...
+pub struct Sha1 {
+    h: [u32; STATE_LEN],
+    block_len: u64,
+    detection: Option<DetectionState>,
+    buffer: BlockBuffer<<Self as BlockSizeUser>::BlockSize, <Self as BufferKindUser>::BufferKind>,
+}
+
+impl HashMarker for Sha1 {}
+
+impl MacMarker for Sha1 {}
```
newpavlov: I don't think this implementation is needed.

sha1/src/checked.rs
```diff
@@ -0,0 +1,386 @@
+//! Collision checked Sha1.
+//!
+//! General techniques and implementation are based on the research and implementation done in [1], [2] by
+//! Marc Stevens.
+//!
+//!
+//! Original license from [3].
+//!
+//! > MIT License
```
newpavlov: I don't think we need to include this licensing information in module-level docs. It can be included behind `//` comments. And `>`s are not needed as well.

newpavlov (commented):
IIUC the checked module does not use anything from the rest of the crate outside of `INITIAL_H` and `STATE_LEN`. Maybe it's better to move this implementation to a separate crate (e.g. `sha1-checked`)? This new crate may also include checked HMAC implementation.

sha1/src/checked.rs
tarcieri: Since we've received permission from @cr-marcstevens to dual license as Apache 2.0+MIT (and per his comment it seems more of the code in the crate is a translation of that code?), I think we can provide some toplevel attribution (e.g. in README.md, which we can import into the rustdoc) mentioning that it contains translations of that code, the original copyright information, and that the resulting translation has been dual licensed Apache 2.0+MIT with permission.

dignifiedquire:
> Maybe it's better to move this implementation to a separate crate

true, should I just add it as a new crate in here?

newpavlov:
>should I just add it as a new crate in here?

Yes. Could you also add a recommendation to the `sha1` crate-level docs to use `sha1-checked` if possible?

sha1-checked/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+edition = "2021"
+documentation = "https://docs.rs/sha1-checked"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["crypto", "sha1", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+rust-version = "1.72"
+
+[dependencies]
+digest = "=0.11.0-pre.8"
+sha1 = { version = "=0.11.0-pre.3", features = [], path = "../sha1" }
```
newpavlov: I don't think you need to depend on `sha1`. Simply copy the initialization constant.

The `sha1` changes can be reverted completely (outside of the doc addition).

sha1/src/lib.rs
```diff
@@ -26,12 +27,16 @@ use digest::const_oid::{AssociatedOid, ObjectIdentifier};
 #[cfg(feature = "zeroize")]
 use digest::zeroize::{Zeroize, ZeroizeOnDrop};
 
+#[cfg(feature = "std")]
+extern crate std;
```
dignifiedquire: I have no idea why I need this, but I am unable to run standard tests without this, am I doing sth wrong?

sha1-checked/Cargo.toml
dignifiedquire: I am using `sha1::compress` as fallback when detection is turned off, that's why the dependency exists

sha1/src/lib.rs
tarcieri: What error are you getting without it?

newpavlov: You forgot to use `default-features = false` when you added `sha1` dependency, since `std` is a default feature, it gets enabled.

sha1/src/checked.rs
dignifiedquire: updated

sha1-checked/Cargo.toml
newpavlov: I think we simply can make collision detection mandatory for this implementation. If users selected this crate over `sha1`, it means they intentionally want to have collision detection.

@tarcieri WDYT?

sha1/src/lib.rs
dignifiedquire: when I remove it from both `sha1` and `sha1-checked` I get this

```
cargo test -p sha1-checked
   Compiling sha1 v0.11.0-pre.3 (/Users/dignifiedquire/opensource/rustcrypto/hashes/sha1)
   Compiling sha1-checked v0.11.0-pre.3 (/Users/dignifiedquire/opensource/rustcrypto/hashes/sha1-checked)
error[E0433]: failed to resolve: use of undeclared crate or module `std`
   --> sha1-checked/src/lib.rs:257:6
    |
257 | impl std::io::Write for Sha1 {
    |      ^^^ use of undeclared crate or module `std`
```

newpavlov: For this you indeed need to use
```rust
#[cfg(feature = "std")]
extern crate std;
```
in `sha1-compressed`. We have [similar lines](https://github.com/RustCrypto/traits/blob/master/crypto-common/src/lib.rs#L12-L13) in other crates.

sha1-checked/Cargo.toml
dignifiedquire: I personally would prefer this setup over having to write a helper function when needing to configure this at runtime, but I am okay either way.

sha1/src/checked.rs
dignifiedquire: removed

dignifiedquire:
I think someone has to approve the new workflows, not sure, might also only get activated when merged

newpavlov:
It's just takes a significant time for them to run with `Cargo.lock` changes. It triggers jobs for all crates, so we have to wait. You also could squash commits into one and force push to cancel some pending jobs.

dignifiedquire:
> You also could squash commits into one and force push to cancel some pending jobs.

squashed a bunch, hopefully this helps cleanup the queue

dignifiedquire:
how far out is a release of `digest@0.11`? I am wondering if it makes sense to backport the current implementation, and publish a `0.10` version, to allow for quick integration, as some projects are already waiting on this

cr-marcstevens:
Before release and deployment, may I ask what kind of testing has been done for whether it is a faithful translation?
For one, I would suggest comparing ubccheck output against the original version for random inputs: this is a carefully crafted filter that determines which full checks can be dismissed a priori, hence an error in translation here could dismiss attacks erroniously before the full check.
And testing sha1 on random inputs and known collisions.

dignifiedquire:
In terms of testing the following has been done
- tests from here https://github.com/cr-marcstevens/sha1collisiondetection/blob/master/Makefile#L111-L120 have been ported for the known collisions, as well as checking for behaviour for these under various conditions
- we are running against all test vectors of the existing sha1 implementation

The `ubc_check` code specifically has not been hand ported, it was translated to rust using c2rust and then only cleaned up carefully, with no logic or constant changes

dignifiedquire:
@tarcieri @newpavlov anything I can help with to move this forward?

newpavlov:
I will try to take a deeper look at it during this weekend. I will merge it right away, if there are no issues.

dignifiedquire:
Thanks, appreciated

newpavlov:
Re: digest v0.11. We do not have an exact date, so it indeed may be worth to release `sha1-checked` based on digest v0.10 first. Luckily, required changes should be relatively small.

dignifiedquire:
Sounds good, I'll change this PR to be based on `digest@0.10` then for now, and then once that is published we can update it back to `0.11`

sha1-checked/Cargo.toml
```diff
@@ -0,0 +1,32 @@
+[package]
+name = "sha1-checked"
+version = "0.11.0-pre.3"
```
newpavlov: This could be simply "0.11" (or v0.10 assuming the `digest` downgrade).

sha1-checked/src/lib.rs
```diff
@@ -0,0 +1,386 @@
...
+    buffer: BlockBuffer<<Self as BlockSizeUser>::BlockSize, <Self as BufferKindUser>::BufferKind>,
+}
+
+impl HashMarker for Sha1 {}
+
+impl BlockSizeUser for Sha1 {
+    type BlockSize = U64;
+}
+
+impl BufferKindUser for Sha1 {
```
newpavlov: `BlockSizeUser` and `BufferKindUser` impls are not necessary and can be removed. These traits are mainly used by the wrapping types in the `digest` crate. The former is also used for HMAC, but I think we will need a separate "checked" HMAC-SHA1 implementation.

sha1-checked/tests/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+        hex!("9ed5d77a4f48be1dbf3e9e15650733eb850897f2"),
+        None,
+        false,
+    );
+}
+
+#[test]
+fn shattered_1() {
+    collision_test(
+        &include_bytes!("./data/shattered-1.pdf")[..],
```
newpavlov: The PDFs are relatively big (almost half MB), so it may be worth to move these tests into a separate file and exclude it together with the PDFs from published package.

sha1-checked/src/compress.rs
```diff
@@ -0,0 +1,791 @@
...
+#[inline]
+pub(super) fn compress(
+    state: &mut [u32; 5],
+    ctx: &mut DetectionState,
+    blocks: &[[u8; BLOCK_SIZE]],
+) {
+    let mut block_u32 = [0u32; BLOCK_SIZE / 4];
+    // since LLVM can't properly use aliasing yet it will make
+    // unnecessary state stores without this copy
+    let mut state_cpy = *state;
```
newpavlov: Since we have a lot of code here, it's very unlikely that we will be able to keep the state in registers, so we can work with the state directly without making a copy.

newpavlov (commented):
I have only minor nits. After those are fixed and `digest` is downgraded to v0.10, I think we can merge it.

sha1-checked/tests/mod.rs
dignifiedquire: I agree, though I had some issues with users complaining they can't run the tests on the version from crates.io

newpavlov: Yes, it's important to keep tests working in published packages. So you need to exclude both the PDFs and `shattered_pdfs.rs` Rust file which uses those.

sha1-checked/src/compress.rs
dignifiedquire: removed

sha1-checked/src/lib.rs
dignifiedquire: removed

sha1-checked/Cargo.toml
dignifiedquire: moved to `0.11.0`, doing a new commit for `0.10`

sha1-checked/tests/mod.rs
cr-marcstevens: Note that the shattered pdfs are already colliding after 320 bytes if i'm not mistaken, and identical after those 320 bytes. So for the purposes of testing it is sufficient to test this colliding prefix of the shattered pdfs.

sha1-checked/Cargo.toml
dignifiedquire: backported to `0.10` and switched to `0.10.0`

sha1-checked/tests/mod.rs
dignifiedquire: Excluded the files, and skipping the test if the file isn't available. Not perfect, but was easier than splitting the tests over multiple files.

wiktor-k:
> I have only minor nits. After those are fixed and `digest` is downgraded to v0.10, I think we can merge it.

Hi! Is there [anything left](https://github.com/RustCrypto/hashes/pull/566#pullrequestreview-1933413295) to get this merged? Sadly some of us are stuck with SHA-1 and this PR would make it a little bit less... explosive. Thanks for your time!

newpavlov:
Sorry, I forgot to return to this PR. I will merge it and release the crate shortly.

PR #569, opened 2024-03-12
sha2: fix aarch64 implementation being disabled

### LINKED ISSUE #568: sha2: aarch64 acceleration broken on master
#542 removed the sha2-asm and asm feature, which incidentally also made the aarch64 code dead, because it is behind a `#[cfg(all(feature = "asm", target_arch = "aarch64"))]` (https://github.com/RustCrypto/hashes/blob/c38787b7abd8dd10412972941d077df54c95a4bb/sha2/src/sha256.rs#L9) — since that feature is no longer declared in Cargo.toml, it cannot be enabled.

I suspect this isn't intentional, because the looongarch implementation also uses inline assembly but isn't disabled.

enkore:
Fixes #568 

(eref 42d478fb6976)

### DIFF
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -6,7 +6,7 @@ cfg_if::cfg_if! {
         mod soft;
         mod x86;
         use x86::compress;
-    } else if #[cfg(all(feature = "asm", target_arch = "aarch64"))] {
+    } else if #[cfg(target_arch = "aarch64")] {
         mod soft;
         mod aarch64;
         use aarch64::compress;
--- a/sha2/src/sha512.rs
+++ b/sha2/src/sha512.rs
@@ -6,7 +6,7 @@ cfg_if::cfg_if! {
         mod soft;
         mod x86;
         use x86::compress;
-    } else if #[cfg(all(feature = "asm", target_arch = "aarch64"))] {
+    } else if #[cfg(target_arch = "aarch64")] {
         mod soft;
         mod aarch64;
         use aarch64::compress;


tarcieri:
Thanks!

enkore:
works quite well, too: on a GH200 the aarch64.rs code runs at 1.7 GiB/s (versus 0.4 GiB/s for the soft implementation), OpenSSL 3 is just shy of 2 GiB/s.

tarcieri:
We can probably switch from inline ASM back to the intrinsics at this point, given the current MSRV

enkore:
Looks like these only landed in 1.76: https://doc.rust-lang.org/core/arch/aarch64/fn.vsha256su0q_u32.html

tarcieri:
I couldn't even get them to work on 1.76, despite the documentation making it look like they should work: https://github.com/RustCrypto/hashes/actions/runs/8251432049/job/22568457459

tarcieri:
Oh, perhaps the SHA-256 intrinsics were stabilized, but the SHA-512 ones weren't?

enkore:
Looks like it, that's a bit odd.

tarcieri:
Here's the stabilization PR: https://github.com/rust-lang/stdarch/pull/1399/files

I don't even see where e.g. `vsha512h2q_u64` is defined? Edit: aha, here it is:

https://github.com/rust-lang/stdarch/blob/3961d3b/crates/core_arch/src/aarch64/neon/generated.rs#L15274-L15288

Gated under `stdarch_neon_sha3`. Tracking issue here: https://github.com/rust-lang/rust/issues/117225

tarcieri:
Opened #571 to track this

PR #570, opened 2024-03-12
sha2: use stabilized `core::arch::aarch64` intrinsics

tarcieri:
We were previously using inline assembly to "emulate" these intrinsics since the ones in `core::arch` had not yet been stabilized.

Given the new 1.72 MSRV, we should now be able to use the intrinsics.


tarcieri:
My bad, I guess these aren't stabilized yet

PR #573, opened 2024-03-15
Add Customized trait for CSHAKE

sylvainpelissier:
Use the new function `new_customized` from https://github.com/RustCrypto/traits/pull/1334 in CSHAKE tests

### DIFF
--- a/sha3/src/macros.rs
+++ b/sha3/src/macros.rs
@@ -469,11 +469,6 @@ macro_rules! impl_cshake {
         pub type $full_name = CoreWrapper<$name>;
 
         impl $name {
-            /// Creates a new CSHAKE instance with the given customization.
-            pub fn new(customization: &[u8]) -> Self {
-                Self::new_with_function_name(&[], customization)
-            }
-
             /// Creates a new CSHAKE instance with the given function name and customization.
             /// Note that the function name is intended for use by NIST and should only be set to
             /// values defined by NIST. You probably don't need to use this function.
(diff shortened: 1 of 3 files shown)


sha3/src/macros.rs
```diff
@@ -613,5 +613,12 @@ macro_rules! impl_cshake {
 
         #[cfg(feature = "zeroize")]
         impl ZeroizeOnDrop for $reader {}
+
+        impl CustomizedInit for $name {
+            #[inline]
+            fn new_customized(customization: &[u8]) -> Self {
+                Self::new(customization)
```
newpavlov: We probably want to remove the inherent method and move its implementation to this trait impl.

sylvainpelissier: I'm not sure to understand how to do that. You meant to move the trait implementation outside of the macro ?

newpavlov: I mean remove the inherent method on lines 443-446 and use `Self::new_with_function_name` in the trait impl instead.

sylvainpelissier: Ah yes, I think it is done now.

newpavlov:
Note that you can patch `digest` in the workspace `Cargo.toml` and use the repository version instead of waiting for a new pre-release.

baloo:
The patch should not be required anymore, if you could rebase this PR.

baloo (approved):

PR #579, opened 2024-05-11
sha: make compress consume blocks

baloo:
To align with the semantics of `block_buffer::BlockBuffer::digest_blocks` signature which works with `&[Block]` and not `&[[u8; N]]`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -16,7 +16,7 @@ name = "ascon-hash"
 version = "0.3.0-pre"
 dependencies = [
  "ascon",
- "digest 0.11.0-pre.8",
+ "digest",
  "hex",
  "hex-literal",
  "spectral",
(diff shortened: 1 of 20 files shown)


newpavlov:
We intentionally use byte arrays in compression functions for two reasons:
- To prepare ground for future migration to const generics (plus, it's useful for `hybrid-array` as well).
- To not deal with the somewhat annoying `Block` imports everywhere.

The compression functions are usually fixed-sized, so I don't think it's worth to expose them to types introduced for dealing with genericity.

baloo:
> * To prepare ground for future migration to const generics (plus, it's useful for `hybrid-array` as well).

Just to be clear, this was post `hybrid-array` migration. Block is an alias for `hybrid_array::Array` already.

PR #586, opened 2024-05-22
sha2: provide support for `spki::AlgorithmIdentifier`

baloo:
This implements `AlgorithmIdentifier` according to [RFC5754 section 2]

[RFC5754 section 2]: https://www.rfc-editor.org/rfc/rfc5754#section-2

Depends on https://github.com/RustCrypto/traits/pull/1575

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -94,14 +94,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "der"
+version = "0.8.0-pre.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b489fd2221710c1dd46637d66b984161fb66134f81437a8489800306bcc2ecea"
+dependencies = [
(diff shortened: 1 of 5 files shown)


baloo:
From the RFC:
```
   The AlgorithmIdentifier parameters field is OPTIONAL.
   Implementations MUST accept SHA2 AlgorithmIdentifiers with absent
   parameters.  Implementations MUST accept SHA2 AlgorithmIdentifiers
   with NULL parameters.  Implementations MUST generate SHA2
   AlgorithmIdentifiers with absent parameters.
```

This implementation generates `AlgorithmIdentifier` with parameters absent, but it would be nice if we could also provide a `PartialEq` that accepted NULL parameters.


(same situation exists in https://www.rfc-editor.org/rfc/rfc3370#section-2.1 for Sha1)

cc @carl-wallace

PR #590, opened 2024-06-10
feat: patch

ratankaliani:
Base for this branch should be `sha2-v.0.9.8` but GitHub doesn't allow for that.

### DIFF
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -157,6 +157,9 @@ cfg_if::cfg_if! {
         mod soft;
         mod aarch64;
         use aarch64::compress;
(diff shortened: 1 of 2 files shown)


sha2/src/sha256.rs
```diff
@@ -157,6 +157,9 @@ cfg_if::cfg_if! {
         mod soft;
         mod aarch64;
         use aarch64::compress;
+    } else if #[cfg(all(target_os = "zkvm", target_vendor = "succinct", target_arch = "riscv32"))] {
```
newpavlov, line 160: Our crates aim for algorithm implementation _in Rust_, so I don't think we will use OS-specific crypto APIs like you did in this PR.

ratankaliani: Mistakenly put these PR's up against the main repo, that makes sense!

PR #591, opened 2024-06-10
Adding the hash Multimixer-128

Anders-Steen:
Did this to exercise my Rust. Any interest having it in the main repo?

### DIFF
--- a/multimixer-128/src/lib.rs
+++ b/multimixer-128/src/lib.rs
@@ -69,7 +69,7 @@ impl MultimixerCore {
 
         for i in 0..4 {
             x[i] = u32::from_ne_bytes([
-                message_block[0 + i * 4],
+                message_block[i * 4],
                 message_block[1 + i * 4],
                 message_block[2 + i * 4],
                 message_block[3 + i * 4],
@@ -105,9 +105,12 @@ impl MultimixerCore {
             p[3] as u64 * q[3] as u64,
         ];
 
-        for i in 0..self.block_sums.len() {
-            self.block_sums[i] = self.block_sums[i].wrapping_add(block_res[i]);
-        }
+        self.block_sums
+            .iter_mut()
+            .zip(block_res.iter())
+            .for_each(|(sum, &res)| {
+                *sum = sum.wrapping_add(res);
+            });
 
         self.block_index += 1;
     }
--- a/multimixer-128/src/lib.rs
+++ b/multimixer-128/src/lib.rs
@@ -15,16 +15,20 @@ use digest::{
     HashMarker, OutputSizeUser, Reset,
 };
 
+const BLOCKSIZE: usize = 32;
+
 pub struct MultimixerCore {
     x: [u32; 4usize],
     y: [u32; 4usize],
     h: [u32; 4usize],
+    k: [u32; 4usize],
     a: [u32; 4usize],
     b: [u32; 4usize],
     p: [u32; 4usize],
     q: [u32; 4usize],
-    key: [u8; 32usize],
+    key_blocks: Vec<Block<Self>>,
     block_sum: [u64; 8usize],
+    block_index: usize,
 }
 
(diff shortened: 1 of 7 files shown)


multimixer-128/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+                h[i] = rng.next_u32();
+                k[i] = rng.next_u32();
+            }
+            (h, k)
+        } else {
+            let mut h = [0u32; 4];
+            let mut k = [0u32; 4];
+            for i in 0..4 {
+                h[i] = u32::from_ne_bytes([
+                    self.key_blocks[self.block_index][i * 4],
```
newpavlov, line 55: What happens with messages bigger than key size? IIUC this code will panic.

multimixer-128/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+use digest::{
+    consts::{U32, U64},
+    core_api::{
+        AlgorithmName, Block, BlockSizeUser, BufferKindUser, CoreWrapper, FixedOutputCore,
+        UpdateCore,
+    },
+    HashMarker, OutputSizeUser,
+};
+use rand_chacha::rand_core::{RngCore, SeedableRng};
+use rand_chacha::ChaCha8Rng;
```
newpavlov, line 16: You should use the `chacha20` crate instead.

multimixer-128/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+
+const BLOCKSIZE: usize = 32;
+
+#[derive(Clone)]
+pub struct MultimixerCore {
+    key_blocks: Vec<Block<Self>>,
+    block_sums: [u64; 8usize],
+    block_index: usize,
+    rng: Option<ChaCha8Rng>,
```
newpavlov, line 25: This looks... suspicious. If you want to make code generic over key blocks generator, I think it's better to make it generic over closure.

multimixer-128/src/lib.rs
```diff
@@ -0,0 +1,224 @@
...
+            rng: Some(ChaCha8Rng::from_seed(
+                key.as_slice()
+                    .try_into()
+                    .expect("Key needs to be able to use as seed."),
+            )),
+        }
+    }
+
+    //Uses key instead of RNG, needs to be same size as message.
+    fn new_from_slice(key: &[u8]) -> Result<Self, InvalidLength> {
```
newpavlov, line 152: In general, `T::new(&key)` and `T::new_from_slice(&key).unwrap()` should produce the same result for the same `key`.

newpavlov (commented):
AFAIK this hash function does not have any real world adoption and it has not received any independent cryptanalysis. The paper has only one citation from the same group. So I am not sure if we want to include this function at the current time.

@tarcieri WDYT?

multimixer-128/src/lib.rs
Anders-Steen: Yes, correct, actually the message and key should be equal in size. I first created the algorithm with a key, but then also implemented a PRNG with a seed. I really wanted to bench the code, and for this I would need a really big key. Ideally I would maybe create two different hash-classes, Multimixer-128 and Multimixer-128-PRNG. I think this would solve the other comments as well.

Anders-Steen:
Thank you for the review! Yes, it should not be used by the public, but as I understood it, maybe as a replacement for the NU hash in [UMAC](https://datatracker.ietf.org/doc/html/rfc4418). I really appreciate the comments, but if it not going to be merged, I think I'll give implementing UMAC a try 🙂. https://github.com/RustCrypto/MACs/issues/1

tarcieri:
Joan Daemen is listed as one of the authors on the paper, so I wouldn't completely discount it, but I'm not seeing a whole lot in the way of peer review on it. It seems very new.

PR #596, opened 2024-06-21
sha1: port `sha1-asm` to `arch::aarch64` intrinsics

### LINKED ISSUE #257: sha1: ARM acceleration using `stdarch` intrinsics
There are a number of issues and comments floating around about this, so I thought I'd make a single tracking issue for this.

Right now we have a separate [asm-hashes](https://github.com/RustCrypto/asm-hashes/) repo containing assembly implementations of various hash functions. 

Longer-term, it would be nice to have ARM implementations of hash functions using [stdarch intrinsics](https://github.com/rust-lang/stdarch/), particularly if they provide similar performance via use of e.g. ARMv8 Cryptography Extensions. I was [almost able prototype](https://github.com/RustCrypto/asm-hashes/issues/5#issuecomment-812890980) a SHA-256 implementation this way:

https://gist.github.com/tarcieri/414a3300072160f372b5d93ccfce280b

There are a few notable blockers though:

- [x] some of the intrinsics we'd need are missing, e.g. `vst1q_u32` and ~~`vreinterpretq_u8_u32`~~ (added)
- [x] ~~the `stdarch` intrinsics for ARM are unstable/nightly-only~~
- [x] ~~there is presently no CI solution for e.g. Apple M1 (we can and already do use `cross` for more generic `aarch64`)~~
- [x] some way of gating the use of these intrinsics, e.g. via `target_feature`/`RUSTFLAGS` and/or runtime detection. see also https://github.com/RustCrypto/utils/issues/378

I think we could move forward prototyping things like SHA-1 and SHA-256 using the `stdarch` intrinsics for the ARMv8 Cryptography Extensions, but we may just need to leave those as draft PRs for now, or if we do merge them potentially add something like a `nightly` feature to gate them under.

newpavlov:
Closes #257

### DIFF
--- a/sha1/src/compress.rs
+++ b/sha1/src/compress.rs
@@ -4,7 +4,7 @@ cfg_if::cfg_if! {
     if #[cfg(feature = "force-soft")] {
         mod soft;
         use soft::compress as compress_inner;
-    } else if #[cfg(all(feature = "asm", target_arch = "aarch64"))] {
+    } else if #[cfg(all(target_arch = "aarch64"))] {
         mod soft;
         mod aarch64;
         use aarch64::compress as compress_inner;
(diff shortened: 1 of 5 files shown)


newpavlov:
@tarcieri 
Could you double check and test that the intrinsic implementation is correct? It passes CI, but it's better to manually check it to be safe.

PR #597, opened 2024-06-27
Added support for kupyna512 hash

AnarchistHoneybun:


kupyna512/src/sub_units/t_xor_plus/tests/test_individual_layers.rs
```diff
@@ -0,0 +1,146 @@
+use crate::sub_units::t_xor_plus::*;
+
+#[test]
+fn test_add_constant_xor() {
+    let input = [
+        0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0A, 0x0B, 0x0C, 0x0D, 0x0E,
```
newpavlov, line 6: You can use the [`hex-literal`](https://docs.rs/hex-literal/latest/hex_literal/) crate for such constants (add it to `dev-dependencies`).

kupyna512/src/sub_units/t_xor_plus/mod.rs
```diff
@@ -0,0 +1,161 @@
...
+    for i in 0..ROWS {
+        for j in 0..COLS {
+            matrix[i][j] = block[i * COLS + j];
+        }
+    }
+    matrix
+}
+
+fn matrix_to_block(matrix: Matrix) -> Vec<u8> {
+    let mut block = vec![0u8; ROWS * COLS];
```
newpavlov, line 27: Heap allocation is not necessary here. You can use byte array instead.

newpavlov (commented):
Please migrate this code to the `digest` traits (see other crates in this repository for examples). The final crate should not contain any heap allocations and should be `no_std`-compatible (i.e. you neeed to add `#![no_std]` to lib.rs).

PR #599, opened 2024-07-03
chore(sha1-checked): remove dead code

### LINKED ISSUE #594: sha1-checked: `ubc_check` is not used
It looks like due to an oversight the `ubc_check` module is not used. Calling the `Builder::use_ubc` has no effect on hash computation.

This results in [CI failures](https://github.com/RustCrypto/hashes/actions/runs/9539622113/job/26290188165).

cc @dignifiedquire

dignifiedquire:
Closes #594


newpavlov (approved):

newpavlov:
Thank you!

PR #601, opened 2024-07-08
Kupyna hasher

jkoudys:
We've implemented this hashing algorithm for kupyna, a Ukrainian algo specified here: https://eprint.iacr.org/2015/885.pdf

I'm opening this PR just to get the ball rolling. The hash function we've made works and we've included a bunch of test units. Next step is to implement the Digest and associated RustCrypto traits. I don't think this should go into RustCrypto/hashes master branch, but figured I'd open this and you could setup a new branch (or let us know the best way to build this). Any tips you could give to get us started on impl'ing  the Digest trait would be  appreciated.

Is there an existing hash implementation with a variable size hash? Most seem to hardcode it, but kupyna allows from 8 to 512, so it ought to be part of its digest struct. Is there a precedent for this already? I'd like to avoid reinventing the wheel.


newpavlov:
See comments in #597.

>Is there an existing hash implementation with a variable size hash?

The [`groestl`](https://github.com/RustCrypto/hashes/blob/master/groestl/src/lib.rs) crate probably will be the easiest to read. You need to implement the [`VariableOutputCore`](https://docs.rs/digest/latest/digest/core_api/trait.VariableOutputCore.html) trait and define appropriate type aliases.

jkoudys:
Nice, thanks this helps.

Can we merge this into a new wip branch, or just get it using a basic Digest and merge to master?

newpavlov:
@jkoudys 
Hi! Do you plan to work on this? As noted in the linked comment, we can not merge it in the current form.

jkoudys:
Yes had a delay but I'm using the groetsl as a guide and putting the kupyna hashing logic around that now.

newpavlov:
Hi! Do you plan to work on this PR further? If so, please take a look at comments in #621.

jkoudys:
@newpavlov I'll close this one. We're working together in #621.

PR #603, opened 2024-07-17
Add missing tests in sha2

CXWorks:
Hi,

Thanks for your time & patience to review this PR.

By examine the existing code, I found some unit tests(7 in total) can be added to improve the repo `sha2` module's overall unit test coverage(this project is already been well tested). Each of the unit test is testing a different target function for better coverage.

Thanks again for reviewing.

### DIFF
--- a/sha2/src/sha512/x86.rs
+++ b/sha2/src/sha512/x86.rs
@@ -356,6 +356,7 @@ const SHA512_HASH_BYTE_LEN: usize = 64;
 const SHA512_HASH_WORDS_NUM: usize = SHA512_HASH_BYTE_LEN / size_of::<u64>();
 const SHA512_BLOCK_WORDS_NUM: usize = SHA512_BLOCK_BYTE_LEN / size_of::<u64>();
 
+#[cfg(target_feature = "avx2")]
 #[cfg(test)]
 mod tests {
     use crate::sha512::x86::sha512_compress_x86_64_avx2;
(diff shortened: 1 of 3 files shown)


sha2/src/sha512/x86.rs
```diff
@@ -355,3 +355,64 @@ const SHA512_ROUNDS_NUM: usize = 80;
 const SHA512_HASH_BYTE_LEN: usize = 64;
 const SHA512_HASH_WORDS_NUM: usize = SHA512_HASH_BYTE_LEN / size_of::<u64>();
 const SHA512_BLOCK_WORDS_NUM: usize = SHA512_BLOCK_BYTE_LEN / size_of::<u64>();
+
+#[cfg(test)]
+mod tests {
+    use crate::sha512::x86::sha512_compress_x86_64_avx2;
+
+    #[test]
+    fn test_sha512_compress_x86_64_avx2() {
```
newpavlov, line 365: This is not a correct way to test this module. You should check that SHA-NI is available, either at compile, or run time.

newpavlov (commented):
I am not sure we need these tests to artificially boost test coverage. A proper coverage measurement should merge several runs which trigger different backends, e.g. by using the crate configuration options.

sha2/src/sha512/x86.rs
CXWorks: Thanks for your time and I will manually resovle it.

CXWorks:
> I am not sure we need these tests to artificially boost test coverage. A proper coverage measurement should merge several runs which trigger different backends, e.g. by using the crate configuration options.

Thanks for your comments and good question. Honestly we don't know a good answer for this question or if this project has a testing coverage plan, but we are happy to get different feedbacks

newpavlov:
Closing this PR since the added test do not test much.

PR #607, opened 2024-07-27
sha2: use stabilized `core::arch::aarch64` intrinsics

tarcieri:
We were previously using inline assembly to "emulate" these intrinsics since the ones in `core::arch` had not yet been stabilized.

They are now stable as of Rust 1.79.

Redux of #570.

### DIFF
--- a/.github/workflows/sha2.yml
+++ b/.github/workflows/sha2.yml
@@ -21,7 +21,7 @@ jobs:
   set-msrv:
     uses: RustCrypto/actions/.github/workflows/set-msrv.yml@master
     with:
-        msrv: 1.72.0
+      msrv: 1.79.0
 
   # Builds for no_std platforms
   build:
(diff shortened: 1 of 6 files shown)


.github/workflows/sha2.yml
```diff
@@ -21,7 +21,7 @@ jobs:
   set-msrv:
     uses: RustCrypto/actions/.github/workflows/set-msrv.yml@master
     with:
-        msrv: 1.72.0
+      msrv: 1.79.0
```
tarcieri, line 24: This seems like an aggressive bump for now but I imagine it will be awhile before we actually get a final version of this crate out

sha2/src/sha256/aarch64.rs
```diff
@@ -4,7 +4,7 @@
 
 // TODO: stdarch intrinsics: RustCrypto/hashes#257
 
-use core::arch::{aarch64::*, asm};
+use core::arch::aarch64::*;
```
newpavlov, line 5: The TODO comment on line 5 can be removed.

newpavlov (approved):

sha2/src/sha256/aarch64.rs
tarcieri: Done in e2556d9

PR #612, opened 2024-08-07
blake2: add unkeyed hashing

### LINKED ISSUE #482: Blake2 parameters cannot be used without a key
The _personal_ and _salt_ parameters for Blake2 can only be used in combination with a key because the non-MAC variants don't expose the parameters as input.

I've also tried using the MAC variant with an empty key, but it appears an initial Blake2 round is done with an all-zero key in that case.

Would it be acceptable if we change the signature of the Mac variant so that (either):

1. key is `Option`, or
2. a key of length 0 does not do the initial Blake2 round?

Or would it be considered a strict violation to use a MAC without a key?

threema-theo:
This introduces unkeyed hashing for blake2 as specified in [Section 2.5 of RFC 7693](https://www.rfc-editor.org/rfc/rfc7693.html#section-2.5) states the following:

    The second (little-endian) byte of the parameter block, "kk", specifies the key size in bytes. Set kk = 00 for unkeyed hashing.

I propose to make the key an `Option<&[u8]>`:
```rust
pub fn new_with_salt_and_personal(
    key: Option<&[u8]>, 
    salt: &[u8], 
    persona: &[u8],
) -> Result<Self, InvalidLength>
```

By making the key an `Option<&[u8]>` - rather than opting for the unkeyed usage in case of an empty `key` - we make the unkeyed usage explicit and avoid inadvertent usages.

This closes #482.
See also #509.

### DIFF
--- a/blake2/src/macros.rs
+++ b/blake2/src/macros.rs
@@ -275,7 +275,7 @@ macro_rules! blake2_mac_impl {
             core: $hash,
             buffer: LazyBuffer<<$hash as BlockSizeUser>::BlockSize>,
             #[cfg(feature = "reset")]
-            key_block: Key<Self>,
+            key_block: Option<Key<Self>>,
             _out: PhantomData<OutSize>,
         }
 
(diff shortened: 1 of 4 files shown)


newpavlov (approved):
Thank you!

PR #614, opened 2024-08-21
sha2: RISC-V scalar crypto extension support

newpavlov:
The support is Nightly-only and requires to enable the `sha2_backend` configuration flag with a value equal to `riscv-zknh` or `riscv-zknh-compact`.

The resulting assembly and binary size of the `compress` function (not counting the statics with round constants):
- SHA-256, unrolled: https://rust.godbolt.org/z/177bqKd3h (5280 bytes)
- SHA-256, compact: https://rust.godbolt.org/z/Kzx59bsdP (1308 bytes)
- SHA-512, unrolled: https://rust.godbolt.org/z/ExqqrfE1r (7964 bytes)
- SHA-512: compact: https://rust.godbolt.org/z/z41v6d4do (2852 bytes)

### DIFF
--- a/.github/workflows/sha2.yml
+++ b/.github/workflows/sha2.yml
@@ -162,6 +162,9 @@ jobs:
       - run: cross test --package sha2 --target riscv64gc-unknown-linux-gnu
         env:
           RUSTFLAGS: '-Dwarnings --cfg sha2_backend="riscv-zknh" -C target-feature=+zknh'
+      - run: cross test --package sha2 --target riscv64gc-unknown-linux-gnu
+        env:
+          RUSTFLAGS: '-Dwarnings --cfg sha2_backend="riscv-zknh-compact" -C target-feature=+zknh'
 
   riscv32-zknh:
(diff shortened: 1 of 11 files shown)


.github/workflows/sha2.yml
```diff
@@ -146,6 +146,42 @@ jobs:
...
+      - uses: actions/checkout@v4
+      - uses: RustCrypto/actions/cargo-cache@master
+      - uses: dtolnay/rust-toolchain@master
+        with:
+          toolchain: nightly
+          components: rust-src
+      - run: cargo build --target riscv32gc-unknown-linux-gnu -Z build-std
+        env:
+          RUSTFLAGS: '-Dwarnings --cfg sha2_backend="riscv-zknh" -C target-feature=+zknh'
+      - run: cargo build --target riscv32gc-unknown-linux-gnu -Z build-std
```
newpavlov, line 181: Note that we do not run tests on RV32 because `cross` currently does not support the `riscv32gc-unknown-linux-gnu` target.

PR #617, opened 2024-08-26
sha2: improve RISC-V Zknh backends

newpavlov:
Annoyingly, RISC-V is really inconvenient when we have to deal with misaligned loads/stores. LLVM by default generates very [inefficient code](https://rust.godbolt.org/z/3Yaj4fq5o) which loads every byte separately and combines them into a 32/64 bit integer. The `ld` instruction "may" support misaligned loads and for Linux user-space it's even [guaranteed](https://www.kernel.org/doc/html/v6.10/arch/riscv/uabi.html#misaligned-accesses), but it can be (and IIUC often in practice is) "extremely slow", so we should not rely on it while writing performant code.

After asking around, it looks like this mess is here to stay, so we have no choice but to work around it. To do that this PR introduces two separate paths for loading block data: aligned and misaligned. The aligned path should be the most common one. In the misaligned path we have to rely on inline assembly since we have to load some bits outside of the block. 

Additionally, this PR makes inlining in the `riscv-zknh` backend less aggressive, which makes generated binary code 3-4 times smaller at the cost of one additional branch.

Generated assembly for RV64:
- SHA-256, unrolled: https://rust.godbolt.org/z/GxPM8PE3P (2278 bytes)
- SHA-256, compact: https://rust.godbolt.org/z/4KWrcve9E (538 bytes)
- SHA-512, unrolled: https://rust.godbolt.org/z/Th8ro8Tbo (2278 bytes)
- SHA-512: compact: https://rust.godbolt.org/z/dqrv48ax3 (530 bytes)

### DIFF
--- a/sha2/src/sha512/riscv_zknh_utils.rs
+++ b/sha2/src/sha512/riscv_zknh_utils.rs
@@ -1,8 +1,16 @@
 use core::{arch::asm, ptr};
 
-#[cfg(target_arch = "riscv32")]
 #[inline(always)]
-pub(super) fn load_aligned_block(block: &[u8; 128]) -> [u64; 16] {
+pub(super) fn load_block(block: &[u8; 128]) -> [u64; 16] {
+    if block.as_ptr().cast::<usize>().is_aligned() {
+        load_aligned_block(block)
+    } else {
+        load_unaligned_block(block)
+    }
+}
+
+#[cfg(target_arch = "riscv32")]
+fn load_aligned_block(block: &[u8; 128]) -> [u64; 16] {
     let p: *const [u32; 32] = block.as_ptr().cast();
     debug_assert!(p.is_aligned());
     let block = unsafe { &*p };
@@ -16,8 +24,7 @@ pub(super) fn load_aligned_block(block: &[u8; 128]) -> [u64; 16] {
 }
 
 #[cfg(target_arch = "riscv64")]
-#[inline(always)]
-pub(super) fn load_aligned_block(block: &[u8; 128]) -> [u64; 16] {
+fn load_aligned_block(block: &[u8; 128]) -> [u64; 16] {
     let block_ptr: *const u64 = block.as_ptr().cast();
     debug_assert!(block_ptr.is_aligned());
     let mut res = [0u64; 16];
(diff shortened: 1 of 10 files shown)


sha2/src/sha512/riscv_zknh_utils.rs
```diff
@@ -0,0 +1,124 @@
...
+    let off2 = (64 - off1) % 64;
+    let bp: *const u64 = block.as_ptr().wrapping_sub(offset).cast();
+
+    let mut left: u64;
+    let mut res = [0u64; 16];
+
+    unsafe {
+        asm!(
+            "ld {left}, 0({bp})",
+            "srl {left}, {left}, {off1}",
```
newpavlov: This `asm!` block (and several others like it) read bits outside of the `block` (e.g. here we read `off1` bits before `block`). In my understanding, this is not an UB. All bits are guaranteed to reside on the same page as at least one byte of the block, meaning that page faults should be impossible on such edge loads. The garbage bits get eliminated by the shift which is intentionally part of the assembly block, so outside of this asm block we can observe only bits which belong to `block`.

UPD: [IRLO discussion](https://internals.rust-lang.org/t/21439)

nazar-pc (commented):
https://github.com/llvm/llvm-project/issues/150263 led me to this PR (the link there wasn't permalink, so it took some effort to chase the PR).

Would it be worth adding a separate case and simplifying block loads when `+unaligned-scalar-mem` feature is specified explicitly?

I have a virtual target that supports this and LLVM already produces A LOT more compact code when this feature is enabled, so not needing to do extra alignment checks + even less code should be even more beneficial. I do use `Zknh` support in `sha2` crate specifically.

newpavlov:
I plan to completely remove the unaligned load hack in a future release and blame LLVM and RISC-V spec for terrible codegen with default compilation options.

nazar-pc:
Oh, does that also mean `+unaligned-scalar-mem` will naturally result in more efficient code in that case by accident? I see why it is unpleasant to maintain such hacks downstream.

newpavlov:
I wouldn't say "by accident", but yes.

newpavlov:
@nazar-pc 
See https://github.com/RustCrypto/hashes/pull/879

PR #635, opened 2024-12-25
added md6

truthixify:
added md6 implementation in pure rust


newpavlov:
Please create a proper PR with implementation code. This PR contains only an incorrectly formed submodule.

truthixify:
How do you suggest I do that?
This is my first time contributing to open source.

newpavlov:
You need to create a branch with committed code and only then create a PR. In you current state of the [`md6` branch](https://github.com/Truthixify/hashes/tree/md6) you only have a "submodule" which does not point anywhere valid.

For further information, please search for it online.

PR #636, opened 2024-12-25
implementation of md6 in pure rust

truthixify:
implementation of md6 in pure rust.


newpavlov (changes requested):
We can not merge this implementation in such state.

Please, do not disable the Clippy lints, follow the trait API defined in the [`digest`](https://docs.rs/digest) crate, keep the crate `no_std`-capable (i.e. your code should not have any `Vec`s), add a CI workflow config for this crate (you can use the [`md5` file](https://github.com/RustCrypto/hashes/blob/master/.github/workflows/md5.yml) for reference), add the crate to the workspace by modifying the [root Cargo.toml file](https://github.com/RustCrypto/hashes/blob/master/Cargo.toml), and remove the `md6/.gitignore` file.

Our trait APIs are a bit complex (and not documented that well), so I recommend taking a look at other crates in this repository first.

truthixify:
@newpavlov 
I don't know why this is failing: sha2 / riscv64-zknh (pull_request)

newpavlov:
The CI failure is fixed in #637, so you need to rebase to master.

md6/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+edition = "2021"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["crypto", "md6", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+rust-version = "1.81"
+
+[lib]
+name = "md6"
```
newpavlov: These two lines are not needed.

md6/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[features]
+default = ["oid", "std"]
+std = ["digest/std"]
+oid = ["digest/oid"]
+zeroize = ["digest/zeroize"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
newpavlov: `docsrs` flag is passed by default during docs.rs builds and no longer needed. I will remove it from other crates in a separate PR.

md6/src/compress.rs
```diff
@@ -0,0 +1,194 @@
+use crate::consts::*;
+
+const W: usize = MD6_W; // number of bits in a word (64)
```
newpavlov: You can simply rename these constants in the `consts` module.

md6/src/compress.rs
```diff
@@ -0,0 +1,194 @@
...
+const C: usize = MD6_C; // size of compression output in words (16)
+const N: usize = MD6_N; // size of compression input block in words (89)
+const Q: usize = MD6_Q; // Q words in a compression block (>= 0) (15)
+const K: usize = MD6_K; // key words per compression block (>= 0) (8)
+const U: usize = MD6_U; // words for unique node ID (0 or 64/w)
+const V: usize = MD6_V; // words for control word (0 or 64/w)
+const B: usize = MD6_B; // data words per compression block (> 0) (64)
+
+const T0: usize = 17; // index for linear feedback
```
newpavlov: I think it's worth to move these constants to the `consts` module.

md6/src/consts.rs
```diff
@@ -0,0 +1,21 @@
+/// MD6 constants related to standard mode of operation
+
+pub(crate) type Md6Word = u64;
+pub(crate) type Md6ControlWord = u64;
+pub(crate) type Md6NodeID = u64;
+
+pub(crate) const MD6_MAX_STACK_HEIGHT: usize = 29; // maximum stack height
```
newpavlov: Please use docstrings for constant descriptions, i.e.:
```rust
/// Maximum stack height
pub(crate) const MD6_MAX_STACK_HEIGHT: usize = 29;
```

md6/src/consts.rs
```diff
@@ -0,0 +1,21 @@
...
+pub(crate) const MD6_MAX_R: usize = 255; // maximum number of rounds
+pub(crate) const MD6_DEFAULT_L: usize = 64; // large so that MD6 is fully hierarchical
+
+pub(crate) const MD6_W: usize = 64; // number of bits in a word
+pub(crate) const MD6_C: usize = 16; // size of compression output in words
+pub(crate) const MD6_N: usize = 89; // size of compression input block in words
+
+/// These five values give lengths of the components of compression
+/// input block; they should sum to MD6_N.
```
newpavlov: It's not a docstring comment. Use `//` instead to comment on a block of constants.

md6/src/consts.rs
```diff
@@ -0,0 +1,21 @@
+/// MD6 constants related to standard mode of operation
+
+pub(crate) type Md6Word = u64;
```
newpavlov: You may just use `pub` for items in this module since you use `pub(crate) mod consts;` in lib.rs.

md6/tests/mod.rs
```diff
@@ -0,0 +1,363 @@
+#![no_std]
```
newpavlov: `#![no_std]` is not needed in tests since tests rely on `std` by default.

Could you also provide a provenance link for these test vectors?

md6/src/md6.rs
```diff
@@ -0,0 +1,743 @@
...
+            bits,
+            i_for_level,
+        })
+    }
+}
+
+impl Md6VarCore {
+    #[inline]
+    fn init(d: usize) -> Self {
+        //
```
newpavlov: Remove this empty comment line?

newpavlov (commented):
Thank you! I have some surface-level comments for now. I will take a closer look at the algorithm implementation later.

md6/src/md6.rs
truthixify: Ok sir

md6/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+[dev-dependencies]
+digest = { version = "=0.11.0-pre.9", features = ["dev"] }
+hex-literal = "0.4"
+base16ct = { version = "0.2", features = ["alloc"] }
+
+[features]
+default = ["oid", "std"]
+std = ["digest/std"]
+oid = ["digest/oid"]
+zeroize = ["digest/zeroize"]
```
newpavlov, line 25: You should keep:
```toml
[package.metadata.docs.rs]
all-features = true
```
Otherwise the generated docs will not contain items gated on the crate features.

md6/src/consts.rs
```diff
@@ -0,0 +1,63 @@
...
+pub const MD6_W: usize = 64;
+/// Size of compression output in words
+pub const MD6_C: usize = 16;
+/// Size of compression input block in words
+pub const MD6_N: usize = 89;
+
+// These five values give lengths of the components of compression
+// input block; they should sum to MD6_N.
+
+// Q words in a compression block (>= 0)
```
newpavlov: Use doc comment (i.e. `///`).

md6/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+version = "0.1.0"
+description = "MD6 hash function"
+authors = ["RustCrypto Developers"]
+license = "MIT OR Apache-2.0"
+readme = "README.md"
+edition = "2021"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["crypto", "md6", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+rust-version = "1.81"
```
newpavlov: Please update to `rust-version = "1.81"` and `edition = "2024"`. Do not forget to update the MSRV badge in the crate README.

md6/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+
+## Minimum Supported Rust Version
+
+Rust **1.81** or higher.
+
+Minimum supported Rust version can be changed in the future, but it will be
+done with a minor version bump.
+
+## SemVer Policy
```
newpavlov: You can remove "Minimum Supported Rust Version" and "SemVer Policy" sections. We decided to relax our MSRV policy after migration to the 2024 edition.

newpavlov:
Could you also add this crate to the algorithms table in the root README?

newpavlov:
Also please add a CI job for this crate. You can use the [md5 config](https://github.com/RustCrypto/hashes/blob/master/.github/workflows/md5.yml) for reference.

md6/Cargo.toml
```diff
@@ -0,0 +1,26 @@
...
+
+[dependencies]
+digest = "=0.11.0-pre.9"
```
newpavlov: Update to `"=0.11.0-pre.10"`.

md6/Cargo.toml
truthixify: You mean rust-version = "1.85"?

newpavlov: Sorry. Yes, I meant 1.85.

.github/workflows/md6.yml
```diff
@@ -0,0 +1,71 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        rust:
+          - ${{needs.set-msrv.outputs.msrv}}
+          - stable
+        target:
+          - thumbv7em-none-eabi
+          - wasm32-unknown-unknown
+          - loongarch64-unknown-linux-gnu
```
newpavlov: `loongarch64-unknown-linux-gnu` can be removed.

md6/Cargo.toml
```diff
@@ -24,3 +24,7 @@ default = ["oid", "std"]
 std = ["digest/std"]
 oid = ["digest/oid"]
 zeroize = ["digest/zeroize"]
+
+[package.metadata.docs.rs]
+all-features = true
+rustdoc-args = ["--cfg", "docsrs"]
```
newpavlov: I meant you should keep `all-features = true`, but we don't need `rustdoc-args = ["--cfg", "docsrs"]`. Also, keep an empty new line at the end of the file.

md6/src/consts.rs
```diff
@@ -18,10 +17,9 @@ pub const MD6_C: usize = 16;
 /// Size of compression input block in words
 pub const MD6_N: usize = 89;
 
-// These five values give lengths of the components of compression
-// input block; they should sum to MD6_N.
-
-// Q words in a compression block (>= 0)
+/// These five values give lengths of the components of compression
+/// input block; they should sum to MD6_N.
```
newpavlov: The docstring was needed only for "Q words in a compression block (>= 0)". These two lines should be kept as usual comments (i.e. as `//`) since they address the block of constants, not just `MD6_Q`.

md6/tests/mod.rs
```diff
@@ -342,12 +342,12 @@ fn test_md6_512() {
...
     ];
 
-    for (msg, &expected_hash) in TEST_VECTOR.iter() {
+    for (msg, expected_hash) in TEST_VECTOR.iter() {
         let mut hasher = md6::Md6_512::new();
         hasher.update(msg);
         let output = hasher.finalize();
 
-        assert!(output.to_vec() == expected_hash);
+        assert!(output.to_vec() == *expected_hash);
```
newpavlov: You don't need to transform output hashes to `Vec`. You can write `assert_eq!(output[..], expected_hash);`.

truthixify: Ok sir, but why do I get this error:

newpavlov: Ah, sorry I haven't checked the suggested line. You can use either `assert_eq!(output[..], expected_hash[..]);` or `assert_eq!(&output[..], expected_hash);`.

(BTW note that you haven't uploaded the image)

md6/src/lib.rs
```diff
@@ -0,0 +1,40 @@
...
+/// Md6 which allows variable output size at runtime
+pub type Md6Var = RtVariableCoreWrapper<Md6VarCore>;
+/// Core hash function for Md6 generic over output size
+pub type Md6Core<OutSize> = CtVariableCoreWrapper<Md6VarCore, OutSize>;
+/// Md6 generic over output size.
+pub type Md6<OutSize> = CoreWrapper<Md6Core<OutSize>>;
+/// Md6 with 64-bit output
+pub type Md6_64 = CoreWrapper<Md6Core<U8>>;
+/// Md6 with 128-bit output
+pub type Md6_128 = CoreWrapper<Md6Core<U16>>;
```
newpavlov: I don't think we need aliases for MD6-64 and MD6-128. In the NIST submission they write the following:
>These are the MD6 versions most relevant for SHA-3: MD6-224, MD6-256, MD6-384, MD6-512.

md6/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+// create a Md6_256 object
+let mut hasher = Md6_256::new();
+
+// write input message
+hasher.update(b"hello world");
+
+// read hash digest and consume hasher
+let hash = hasher.finalize();
+assert_eq!(hash.to_vec(), hex!(
```
newpavlov: Do not use `to_vec()` here as well.

md6/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+keywords = ["crypto", "md6", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+
+[dependencies]
+digest = "=0.11.0-pre.10"
+
+[dev-dependencies]
+digest = { version = "=0.11.0-pre.10", features = ["dev"] }
+hex-literal = "0.4"
+base16ct = { version = "0.2", features = ["alloc"] }
```
newpavlov: You don't use `base16ct` in tests or example, so it can be removed. Alternatively, you could add a hex-encoding example similarly to the other crates.

md6/src/md6.rs
```diff
@@ -0,0 +1,737 @@
...
+    } else if w == 16 {
+        &[
+            0x7311, 0xc281, 0x2425, 0xcfa0, 0x6432, 0x2864, 0x34aa, 0xc8e7, 0xb604, 0x50e9, 0xef68,
+            0xb7c1, 0xe8fb, 0x2390, 0x8d9f, 0x06f1, 0xdd2e, 0x76cb, 0xa691, 0xe5bf, 0x0cd0, 0xd63b,
+            0x2c30, 0xbc41, 0x1f8c, 0xcf68, 0x2305, 0x8f8a, 0x54e5, 0xed5b, 0x88e3, 0x775d, 0x4ad1,
+            0x2aae, 0x0a6d, 0x6031, 0x3e7f, 0x16bb, 0x8822, 0x2e0d, 0x8af8, 0x671d, 0x3fb5, 0x0c2c,
+            0x995a, 0xd117, 0x8bd2, 0x5c31, 0xc878, 0xc1dd, 0x04c4, 0xb633, 0x3b72, 0x066c, 0x7a15,
+            0x52ac, 0x0d6f, 0x3522, 0x631e, 0xffcb,
+        ]
+    } else if W == 8 {
```
newpavlov: I think you should use `w` here instead of `W`?

md6/tests/mod.rs
truthixify: image

truthixify: @newpavlov please check the image here, I've uploaded it.

newpavlov: You need to add a gated import:
```rust
#[cfg(feature = "zeroize")]
use digest::zeroize::{Zeroize, ZeroizeOnDrop};
```

md6/src/consts.rs
```diff
@@ -0,0 +1,61 @@
...
+/// Q words in a compression block (>= 0)
+pub const MD6_Q: usize = 15;
+/// Key words per compression block (>= 0)
+pub const MD6_K: usize = 8;
+/// Words for unique node ID (0 or 64/w)
+pub const MD6_U: usize = 64 / MD6_W;
+/// Words for control word (0 or 64/w)
+pub const MD6_V: usize = 64 / MD6_W;
+/// Data words per compression block (> 0)
+pub const MD6_B: usize = 64;
```
newpavlov: I don't think we need these separate `MD6_*` constants. Just inline the values to `W`, `C`, `N`, etc. constants defined below.

md6/src/consts.rs
```diff
@@ -0,0 +1,61 @@
...
+
+/// Number of bits in a word
+pub const MD6_W: usize = 64;
+/// Size of compression output in words
+pub const MD6_C: usize = 16;
+/// Size of compression input block in words
+pub const MD6_N: usize = 89;
+
+// These five values give lengths of the components of compression
+// input block; they should sum to MD6_N.
```
newpavlov: You can add the following code to statically assert this property:
```rust
const _: () = {
    if Q + K + U + V != N {
        panic!("`Q + K + U + V` must be equal to `N`");
    }
};
```

truthixify: Okay sir.

md6/src/compress.rs
```diff
@@ -0,0 +1,178 @@
...
+    // Get the initial values for `s` and `smask` based on the width `w`.
+    let (mut s, smask) = get_s_constants(W);
+
+    let mut i = N;
+    let mut j = 0;
+
+    while j < r * C {
+        // Call the loop bodies based on the value of `w`.
+        // This will perform the main computation for each step in the compression loop.
+        call_loop_bodies!(W, s, i);
```
newpavlov, line 75: You always use the `W` constant with this macro. Why do you then handle other potential `W` values in the macro?

You either need to change `W` depending on compilation target (it's fine to support only 32 and 64 bit targets), or you need to remove the dead branches.

md6/src/consts.rs
```diff
@@ -10,42 +10,28 @@ pub const MD6_MAX_R: usize = 255;
...
 /// Words for control word (0 or 64/w)
-pub const V: usize = MD6_V;
+pub const V: usize = 1;
 /// Data words per compression block (> 0) (64)
-pub const B: usize = MD6_B;
+pub const B: usize = 64;
+
+const _: () = {
+    if Q + K + U + V + B != N {
+        panic!("`Q + K + U + V` must be equal to `N`");
```
newpavlov: Sorry, forgot `B`, it should be "`Q + K + U + V + B`" in the panic message.

md6/src/compress.rs
truthixify: The reason why I did that was because, I'd like to support compilation target.

newpavlov: Then you need to change the `W` constant depending on target:
```rust
#[cfg(target_pointer_width = "64")]
pub const N: usize = 64;
#[cfg(target_pointer_width = "32")]
pub const N: usize = 32;
```

You can remove branches responsible for 16 and 8 bit targets since Rust compiler does not support those.

truthixify: Okay sir.

truthixify:
Hello sir @newpavlov.

I've added all the changes sir.

PR #652, opened 2025-02-22
Upgrade all crates to the 2024 edition and bump MSRV to 1.85

newpavlov:


jrose-signal:
What motivated this change? It's a more aggressive MSRV update than RustCrypto has done in the past, and I don't see any changes in any of these three linked PRs that *required* it.

tarcieri:
All of the crates already had relatively high MSRV (1.81) for `core::error::Error`, which allows us to drop the `std` feature.

We have very long release cycles, often several years, for which in the past we have maintained MSRV compatibility. The 2024 edition enables the MSRV-aware resolver, which means we don't get locked into artificially low MSRVs for several years.

jrose-signal:
The 2024 edition enables the MSRV-aware resolver *for the root crates in a workspace.* It doesn't change anything about resolution when your crate is a dependency, because (a) picking a version of your crate happens as part of the resolver's work, and (b) the downstream dependent can still be on the 2021 edition, or the 2015 edition, or have manually turned off the new resolver for some reason.

tarcieri:
@jrose-signal look, we just want to bump all of our crates to the 2024 edition, consistently, across this entire project.

We could nickle-and-dime each of the hundreds of our crates and figure out the exact minimal MSRV each one needs treating each crate as a unique snowflake, but it's *much* easier to just do everything consistently.

Ask yourself if you will really care whether the MSRV is 1.81 or 1.85 in 3 months.

(Note: I am having many arguments about MSRV across several projects which are quite frankly pushing me towards burnout and ragequitting OSS altogether, so also ask yourself if you really want to burn out the project maintainers trying to gatekeep what's an appropriate MSRV and how much that means to you)

Issue #546, opened 2024-01-11
jh: remove dependency on `ppv-lite86`


newpavlov:
Depending on `ppv-lite86` by default makes it harder to support `no_std` and `zeroize`. Ideally we should rewrite the compressor using arch intrinsics and `cpufeatures`.

It also worth to implement the `SerializableState` trait for consistency with other crates.

PR #678, opened 2025-04-07
Migrate to buffering macros

newpavlov:
Depends on https://github.com/RustCrypto/traits/pull/1799.


sha3/src/cshake.rs
```diff
@@ -0,0 +1,200 @@
...
+                }
+            }
+        }
+
+        #[cfg(feature = "zeroize")]
+        impl digest::zeroize::ZeroizeOnDrop for $name {}
+
+        impl SerializableState for $name {
+            // TODO: implement properly with U400
+            type SerializedStateSize = digest::consts::U0;
```
newpavlov: To properly implement this trait we need to implement optimized serialization for eager hashes (otherwise we would need to implement `ArraySize` for `U537` and `U569`). I plan to do it in later PRs.

jh/src/block_api.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[cfg(feature = "zeroize")]
+impl digest::zeroize::ZeroizeOnDrop for JhCore {}
+
+impl SerializableState for JhCore {
+    type SerializedStateSize = U64;
+
+    fn serialize(&self) -> SerializedState<Self> {
+        todo!()
+    }
```
newpavlov: This is a placeholder impl. To implement it properly we probably need to resolve #546 first.

sha1/README.md
```diff
@@ -9,7 +9,7 @@
 
 Pure Rust implementation of the [SHA-1] cryptographic hash algorithm.
 
-## 🚨 Warning: Cryptographically Broken! 🚨
+## Warning: Cryptographically Broken!
```
newpavlov, line 12: The emojis are removed because they cause ICE on the latest Nightly. We can return them later after the bug is fixed.

PR #683, opened 2025-04-24
Use a const lookup table for the kupyna gf

jkoudys:
The GF_LOOKUP_TABLE was being computed on each and every call. This was quite slow, but  all but the most trivial hashes are going to iterate the full 256x256 values anyway. Rust has the beautiful `const fn` syntax, which lets us easily adapt the old runtime code with few changes aside from using a `while` instead of a `for` inside.

The kupyna algo still has a lot of optimizing before it hits the same performance as the similar groetsl algo, but this simple change sees a major improvement.

### DIFF
--- a/kupyna/src/utils.rs
+++ b/kupyna/src/utils.rs
@@ -120,3 +120,4 @@ pub(crate) fn write_u64_le(src: &[u64], dst: &mut [u8]) {
         dst.copy_from_slice(&src.to_le_bytes())
     }
 }
+
--- a/kupyna/src/utils.rs
+++ b/kupyna/src/utils.rs
@@ -120,4 +120,3 @@ pub(crate) fn write_u64_le(src: &[u64], dst: &mut [u8]) {
         dst.copy_from_slice(&src.to_le_bytes())
     }
 }
-
--- a/kupyna/src/utils.rs
+++ b/kupyna/src/utils.rs
@@ -1,14 +1,13 @@
 use crate::consts::{MDS_MATRIX, SBOXES};
 
 const fn gf_multiply(x: u8, y: u8) -> u8 {
-    const BITS_IN_BYTE: u8 = 8;
     const REDUCTION_POLYNOMIAL: u16 = 0x011d;
 
     let mut x = x;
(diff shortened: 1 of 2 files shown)


kupyna/src/utils.rs
```diff
@@ -16,11 +18,32 @@ fn multiply_gf(mut x: u8, mut y: u8) -> u8 {
...
+        while y < 256 {
+            table[x][y] = gf_multiply(x as u8, y as u8);
+            y += 1;
+        }
+        x += 1;
+    }
+    table
+}
+
+pub(crate) static GF_LOOKUP_TABLE: [[u8; 256]; 256] = generate_gf_lookup_table();
```
newpavlov: I think you can keep this static private since it's only used by the `multiply_gf` function.

kupyna/src/utils.rs
```diff
@@ -16,11 +18,32 @@ fn multiply_gf(mut x: u8, mut y: u8) -> u8 {
...
+
+fn multiply_gf(x: u8, y: u8) -> u8 {
+    GF_LOOKUP_TABLE[x as usize][y as usize]
```
newpavlov: You can use `usize::from(x)` instead of `as`.

kupyna/src/utils.rs
```diff
@@ -1,12 +1,14 @@
...
     const BITS_IN_BYTE: u8 = 8;
     const REDUCTION_POLYNOMIAL: u16 = 0x011d;
 
+    let mut x = x;
+    let mut y = y;
     let mut r = 0u8;
-
-    for _ in 0..BITS_IN_BYTE {
+    let mut i = 0;
+    while i < BITS_IN_BYTE {
```
newpavlov: You can use `u8::BITS` instead of using manually defined `BITS_IN_BYTE`.

newpavlov (commented):
I have several nitpicks, but otherwise looks good!

jkoudys:
I enjoy picking a good nit. All resolved.

newpavlov (approved):

PR #686, opened 2025-04-28
sha2: add soft-compact backend

### LINKED ISSUE #561: soft-sha512 code size seems unreasonably high on thumbv7em
I'm working on an embedded project which needs ed25519 signing, which pulls in sha2 for the sha512 step of the signing procedure. On an `thumbv7em` target, the sha512 implementation appears to consume a very significant chunk of code space:

```
$ cargo bloat --release --filter sha2
File .text    Size Crate Name
0.9% 17.0% 26.8KiB  sha2 sha2::sha512::compress512
0.2%  4.4%  6.9KiB  sha2 sha2::sha256::compress256
0.0%  0.0%      0B       And 0 smaller methods. Use -n N to show more.
1.1% 21.4% 33.7KiB       filtered data size, the file size is 3.0MiB
```

If I use `opt-level = "z"` it is better, but still pretty high:

```
$ cargo bloat --release --filter sha2
File .text    Size Crate Name
0.5% 10.1% 10.8KiB  sha2 sha2::sha512::compress512
0.2%  3.0%  3.2KiB  sha2 sha2::sha256::compress256
0.0%  0.2%    250B  sha2 sha2::sha512::soft::sha512_schedule_x2
0.0%  0.2%    210B  sha2 sha2::sha512::soft::sha512_digest_round
0.0%  0.1%    162B  sha2 sha2::sha256::soft::schedule
0.0%  0.1%    162B  sha2 sha2::sha256::soft::sha256_digest_round_x2
0.0%  0.0%     32B  sha2 core::iter::adapters::zip::TrustedRandomAccessNoCoer...
0.0%  0.0%      0B       And 0 smaller methods. Use -n N to show more.
0.7% 13.8% 14.8KiB       filtered data size, the file size is 2.0MiB
```

The amounts above seem quite onerous for my target which only has 256kB of flash, and would potentially be a non-starter for targets with even less code storage available.

I suspect it's probably a combination of inlined soft 64-bit integer arithmetic and extreme levels of code generation due to macro expansion.

If improving the code size would be a performance regression on some platforms perhaps an implementation favoring code size gated behind a crate feature flag could be of interest? Like a "32-bit-friendly" version or something I guess.

newpavlov:
The new backend results in a much smaller binary code (e.g. for SHA-256 on `thumbv7em-none-eabi` we get [~150 instructions][0] vs [~1680 instructions][1], for SHA-512 it's [~330][2] vs [~7950][3]) while being somewhat slower on non-embedded CPUs (e.g. for SHA-256 on my x86-64 laptop with `opt-level=s` it results in 289 MB/s vs 315 MB/s).

We probably should also backport this PR to the `backports/digest-v0.10` branch and release sha2 v0.10.9 with this addition.

Closes #561

[0]: https://rust.godbolt.org/z/c5cT5chGz
[1]: https://rust.godbolt.org/z/rWMTqaxdc
[2]: https://rust.godbolt.org/z/EKoxxdqrv
[3]: https://rust.godbolt.org/z/Td81PxsdW

### DIFF
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -42,14 +42,10 @@ cfg_if::cfg_if! {
 #[inline(always)]
 #[allow(dead_code)]
 fn to_u32s(block: &[u8; 64]) -> [u32; 16] {
-    let mut res = [0u32; 16];
-    // note: we intentionally do not use `zip`-based code here since
-    // it results in a suboptimal codegen for `opt-level = "s"`
-    for i in 0..16 {
+    core::array::from_fn(|i| {
         let chunk = block[4 * i..][..4].try_into().unwrap();
-        res[i] = u32::from_be_bytes(chunk);
-    }
-    res
+        u32::from_be_bytes(chunk)
+    })
 }
 
 /// Raw SHA-256 compression function.
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -2,6 +2,9 @@ cfg_if::cfg_if! {
     if #[cfg(sha2_backend = "soft")] {
         mod soft;
         use soft::compress;
+    } else if #[cfg(sha2_backend = "soft-compact")] {
+        mod soft_compact;
+        use soft_compact::compress;
     } else if #[cfg(any(target_arch = "x86", target_arch = "x86_64"))] {
         mod soft;
         mod x86_shani;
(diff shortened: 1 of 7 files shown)


sha2/src/sha256.rs
```diff
@@ -40,8 +43,11 @@ cfg_if::cfg_if! {
 #[allow(dead_code)]
 fn to_u32s(block: &[u8; 64]) -> [u32; 16] {
     let mut res = [0u32; 16];
-    for (src, dst) in block.chunks_exact(4).zip(res.iter_mut()) {
-        *dst = u32::from_be_bytes(src.try_into().unwrap());
+    // note: we intentionally do not use `zip`-based code here since
+    // it results in a suboptimal codegen for `opt-level = "s"`
+    for i in 0..16 {
```
newpavlov: This may be just an artifact of how godbolt (or the Rust compiler) generates assembly for `opt-level=s`, see: https://users.rust-lang.org/t/128919 But I will leave this change just to be extra safe.

tarcieri (approved):

PR #693, opened 2025-05-26
refactor: Kupyna - replace byte-level operations with word-level operations

AnarchistHoneybun:
(had to scrap the previous pr due to a botched merge from master)


kupyna/src/long.rs
```diff
@@ -1,95 +1,74 @@
...
-    const COLS: usize = 8;
-
-    let mut matrix = [[0u8; COLS]; ROWS];
-    for i in 0..ROWS {
-        for j in 0..COLS {
-            matrix[i][j] = block[i * COLS + j];
+fn rotate_rows(state: [u64; COLS]) -> [u64; COLS] {
+    // Convert to matrix format (column-major as per paper)
+    let mut matrix = [[0u8; COLS]; 8];
+    for col in 0..COLS {
```
newpavlov: If you want, you can use `#[allow(clippy::needless_range_loop)]` on this ans similar functions.

AnarchistHoneybun: these are the ones with the u64 to u8 conversions, I'd rather figure out a way to bypass that.

AnarchistHoneybun:
feel like from_be_bytes and to_be_bytes in a loop just ends up doing the same thing as the full matrix conversion. tried something different for mix_columns, still benchmarking to see if it does better. working on in-place mods next

AnarchistHoneybun:
the needle finally moved on performance:
```
running 8 tests
test kupyna256_10    ... bench:         323.36 ns/iter (+/- 10.25) = 30 MB/s
test kupyna256_100   ... bench:       3,227.37 ns/iter (+/- 1,752.45) = 30 MB/s
test kupyna256_1000  ... bench:      43,945.53 ns/iter (+/- 1,924.61) = 22 MB/s
test kupyna256_10000 ... bench:     436,109.65 ns/iter (+/- 119,690.41) = 22 MB/s
test kupyna512_10    ... bench:         646.24 ns/iter (+/- 50.07) = 15 MB/s
test kupyna512_100   ... bench:       7,073.20 ns/iter (+/- 2,724.30) = 14 MB/s
test kupyna512_1000  ... bench:      66,458.66 ns/iter (+/- 3,645.49) = 15 MB/s
test kupyna512_10000 ... bench:     678,858.70 ns/iter (+/- 12,993.73) = 14 MB/s
```
it's miniscule, but yeah

kupyna/src/utils.rs
```diff
@@ -44,58 +44,159 @@ fn multiply_gf(x: u8, y: u8) -> u8 {
...
-    round: usize,
-) -> [[u8; 8]; N] {
-    for (j, row) in state.iter_mut().enumerate() {
+pub(crate) fn add_constant_xor<const N: usize>(state: &mut [u64; N], round: usize) {
+    for (j, word) in state.iter_mut().enumerate() {
         let constant = ((j * 0x10) ^ round) as u8;
-        row[0] ^= constant;
+        let mut bytes = word.to_be_bytes();
+        bytes[0] ^= constant;
+        *word = u64::from_be_bytes(bytes);
```
newpavlov: I don't think you need to transform the word to bytes here. Instead you can create a 64 bit constant which would contain the appropriate value in the least significant byte (note that the constant will be different for LE and BE targets).

kupyna/src/utils.rs
```diff
@@ -44,58 +44,159 @@ fn multiply_gf(x: u8, y: u8) -> u8 {
...
+        let transformed_bytes = [
+            SBOXES[0][bytes[0] as usize], // row 0 -> S-box 0
+            SBOXES[1][bytes[1] as usize], // row 1 -> S-box 1
+            SBOXES[2][bytes[2] as usize], // row 2 -> S-box 2
+            SBOXES[3][bytes[3] as usize], // row 3 -> S-box 3
+            SBOXES[0][bytes[4] as usize], // row 4 -> S-box 0 (4 % 4 = 0)
+            SBOXES[1][bytes[5] as usize], // row 5 -> S-box 1 (5 % 4 = 1)
+            SBOXES[2][bytes[6] as usize], // row 6 -> S-box 2 (6 % 4 = 2)
+            SBOXES[3][bytes[7] as usize], // row 7 -> S-box 3 (7 % 4 = 3)
+        ];
```
newpavlov: You can write it like this:
```rust
let transformed_bytes = core::array::from_fn(|i| {
    SBOXES[i % 4][bytes[i] as usize]
});
```

kupyna/src/utils.rs
```diff
@@ -44,58 +44,159 @@ fn multiply_gf(x: u8, y: u8) -> u8 {
...
-    state
 }
 
-pub(crate) fn add_constant_xor<const N: usize>(
-    mut state: [[u8; 8]; N],
-    round: usize,
-) -> [[u8; 8]; N] {
-    for (j, row) in state.iter_mut().enumerate() {
+pub(crate) fn add_constant_xor<const N: usize>(state: &mut [u64; N], round: usize) {
+    for (j, word) in state.iter_mut().enumerate() {
```
newpavlov: Use `i` instead of `j` here?

kupyna/src/utils.rs
```diff
@@ -44,58 +44,159 @@ fn multiply_gf(x: u8, y: u8) -> u8 {
...
-pub(crate) fn add_constant_plus<const N: usize>(
-    mut state: [[u8; 8]; N],
-    round: usize,
-) -> [[u8; 8]; N] {
-    for (j, row) in state.iter_mut().enumerate() {
-        let mut row_as_u64 = u64::from_le_bytes(*row);
+pub(crate) fn add_constant_plus<const N: usize>(state: &mut [u64; N], round: usize) {
+    for (j, word) in state.iter_mut().enumerate() {
+        // Convert to little-endian bytes to match original behavior
+        let mut row_as_u64 = u64::from_le_bytes(word.to_be_bytes());
```
newpavlov: You probably should use `from_be` or `from_le` here.

Also, ideally you should not reverse byte order here in the first place and instead just generate a proper constant for LE and BE targets.

kupyna/src/utils.rs
```diff
@@ -44,58 +44,159 @@ fn multiply_gf(x: u8, y: u8) -> u8 {
...
-    result
+        // Extract each byte (row) from the u64 using bit shifts and masks
+        let byte0 = ((input >> 56) & 0xFF) as u8;
+        let byte1 = ((input >> 48) & 0xFF) as u8;
+        let byte2 = ((input >> 40) & 0xFF) as u8;
+        let byte3 = ((input >> 32) & 0xFF) as u8;
+        let byte4 = ((input >> 24) & 0xFF) as u8;
+        let byte5 = ((input >> 16) & 0xFF) as u8;
+        let byte6 = ((input >> 8) & 0xFF) as u8;
+        let byte7 = (input & 0xFF) as u8;
```
newpavlov: You can simplify this function like this:
```rust
let bytes = word.to_be_bytes();
let transformed_bytes = core::array::from_fn(|i| {
    let t = multiply_gf_array(&bytes, &MDS_MATRIX[i]);
    multiply_gf(bytes[i], t);
});
state[col] = u64::from_be_bytes(transformed_bytes);


fn multiply_gf_array(a: &[u8; 8], b: &[u8; 8]) -> u8 {
    let mut res = 0;
    for i in 0..8 {
        res ^= multiply_gf(a[i], b[i]);
    }
    res
}

```

kupyna/src/long.rs
```diff
@@ -1,95 +1,64 @@
...
 
-    for (dst, src) in prev_vector.iter_mut().zip(prev_vector_u8.chunks_exact(8)) {
-        *dst = u64::from_be_bytes(src.try_into().unwrap());
-    }
+    *prev_vector = xor(xor(t_xor_mp, t_plus_m), *prev_vector);
 }
 
-pub(crate) fn t_plus_l(block: [u8; 128]) -> [u8; 128] {
-    let mut state = block_to_matrix(block);
+pub(crate) fn t_plus_l(state: [u64; COLS]) -> [u64; COLS] {
```
newpavlov: I think this function can be made private.

kupyna/src/block_api.rs
```diff
@@ -72,23 +72,18 @@ impl VariableOutputCore for KupynaShortVarCore {
...
 
-        // Call t_xor_l with u8 array
-        let t_xor_ult_processed_block = short::t_xor_l(state_u8);
+        let result_state = xor(self.state, t_xor_ult_processed_block);
 
-        let result_u8 = xor_bytes(state_u8, t_xor_ult_processed_block);
+        // println!("final_result(pre truncate):=");
+        // for v in result_state.iter() {
+        //     println!("{:02X?}", v);
+        // }
```
newpavlov: Delete these comments.

kupyna/Cargo.toml
```diff
@@ -14,6 +14,7 @@ rust-version = "1.85"
 
 [dependencies]
 digest = "=0.11.0-pre.10"
+hex-literal = "1"
```
newpavlov: Remove this dependency.

newpavlov (commented):
You should be able to improve performance by replacing instances of `to_be_bytes`/`from_be_bytes` with `to_le_bytes`/`from_le_bytes`. For example, compare codegen for these two snippets: https://rust.godbolt.org/z/GMKTvKbe1 and https://rust.godbolt.org/z/qMfe7c4qf Obviously, the snippets are not equiavalent, but we can see that in the second case the compiler is able to generate more efficient SIMD-based code.

newpavlov:
Thank you!

PR #698, opened 2025-06-24
update actions/checkout from v3 to v4

PixelPil0t1:
Updates actions/checkout to the latest stable version v4.

Changes:

Updates all instances of actions/checkout@v3 to actions/checkout@v4

Reference:
Latest version confirmation: https://github.com/actions/checkout/releases/tag/v4.2.2

This update ensures we're using the most recent stable version of the GitHub checkout action across our CI/CD pipelines.

### DIFF
--- a/.github/actions/cross-tests/action.yml
+++ b/.github/actions/cross-tests/action.yml
@@ -13,7 +13,7 @@ inputs:
 runs:
   using: "composite"
   steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


tarcieri:
Surprised Dependabot didn't do this automatically: https://github.com/RustCrypto/hashes/blob/master/.github/dependabot.yml#L8-L12

newpavlov:
Yeah, I am also surprised. I will probably move the job to the actions repo and will update the version as part of it. This PR looks like an automated one, so it can be closed.

PR #700, opened 2025-06-25
Replace dead Tiger hash link with official project page

VolodymyrBg:
The outdated and non-working link to the Tiger hash function specification was replaced with the official project page (https://biham.cs.technion.ac.il/Reports/Tiger/). This update ensures that users have access to the authoritative source for the Tiger algorithm, including its description, reference implementation, and related resources.


newpavlov:
Closing in favor of #701

PR #712, opened 2025-08-01
perf(sha2): optimize SHA-256 memory usage with circular buffer approach

DeVikingMark:
This PR improves the memory efficiency of the SHA-256 implementation by replacing the full 64-element message schedule (w[64]) with a circular buffer approach using the original 16-element block array.

### DIFF
--- a/sha2/src/sha256/soft_compact.rs
+++ b/sha2/src/sha256/soft_compact.rs
@@ -1,29 +1,31 @@
 use crate::consts::K32;
 
-fn compress_u32(state: &mut [u32; 8], block: [u32; 16]) {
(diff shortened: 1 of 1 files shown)


newpavlov:
Have you measured performance impact of this change? You can use the `sha2_backend = "soft"` configuration flag to force use of the software backend.

newpavlov:
On my PC this PR improves performance from ~277 MB/s to ~285 MB/s. It also results in smaller binary code and stack usage.

newpavlov (approved):

newpavlov:
Thank you!

PR #713, opened 2025-08-04
fix: add missing OID for RIPEMD-320 (Teletrust)

0xbryer:


newpavlov:
The OID is not correct and we do not accept automated PRs from first-time contributors.

Issue #718, opened 2025-08-11
groestl: add AVX512/GFNI backend


robbie01:
Last year, I wrote an [AVX512/GFNI implementation](https://github.com/robbie01/groestl-gfni) of the "short" core for Groestl. (Ignore the "long" core code in g512.rs, it doesn't work correctly last I checked). Now that these intrinsics are stabilized, I would love to get this up to the standards of the project and contribute it as an optimized backend.

newpavlov:
Sounds interesting. Feel free to open a PR. But I don't think we should use `(__m512i, __m512i)` for the hash state. Instead we should keep `[u64; COLS]`, use unaligned loads, and change the compress function to operate over slice of blocks instead of one block at a time.

robbie01:
@newpavlov Exactly the kind of feedback I was looking for. I’ll start hashing out a PR. Is there any sort of project-wide standard for introducing selectable backends? I understand that ad-hoc macros seem to be used throughout RustCrypto.

PR #720, opened 2025-08-13
groestl: add AVX-512/GFNI backend

robbie01:
#718

I took a conservative approach here and kept the same in-memory representation as the original (now `soft`) backend. This results in an extra two `vpermb`s (`_mm512_permutexvar_epi8`) per call to `compress` and `p`. (Note: this is *not* a per-block overhead, as `compress` now works on a slice of blocks per @newpavlov's recommendation.)

If it's acceptable, I can modify the code to use the same state representation in memory as it does in the register. It should be risk-free as it would be absurd for CPU features to change during execution.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -134,6 +134,8 @@ name = "groestl"
 version = "0.11.0-pre"
 dependencies = [
  "base16ct",
+ "cfg-if",
+ "cpufeatures",
  "digest",
  "hex-literal",
 ]
--- a/groestl/Cargo.toml
+++ b/groestl/Cargo.toml
@@ -14,6 +14,10 @@ categories = ["cryptography", "no-std"]
 
 [dependencies]
 digest = "0.11.0-rc.0"
+cfg-if = "1"
+
+[target.'cfg(any(target_arch = "x86_64", target_arch = "x86"))'.dependencies]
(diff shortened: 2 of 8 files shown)


robbie01:
Performance (with `-C target-cpu=native`, Ryzen 9 7900X, `x86_64-pc-windows-msvc`):

`soft` backend:
```
test groestl256_10    ... bench:          62.62 ns/iter (+/- 1.15) = 161 MB/s
test groestl256_100   ... bench:         604.86 ns/iter (+/- 7.71) = 165 MB/s
test groestl256_1000  ... bench:       5,930.86 ns/iter (+/- 83.92) = 168 MB/s
test groestl256_10000 ... bench:      59,241.11 ns/iter (+/- 535.22) = 168 MB/s
```

`avx512_gfni` backend:
```
test groestl256_10    ... bench:          15.39 ns/iter (+/- 0.42) = 666 MB/s
test groestl256_100   ... bench:         148.98 ns/iter (+/- 5.03) = 675 MB/s
test groestl256_1000  ... bench:       1,402.30 ns/iter (+/- 27.58) = 713 MB/s
test groestl256_10000 ... bench:      13,936.83 ns/iter (+/- 608.29) = 717 MB/s
```

PR #733, opened 2025-09-09
fix: replace removed sha2 compression function exports

### LINKED ISSUE #732: sha2 `compress256` no longer public
After https://github.com/RustCrypto/hashes/pull/678, `sha2::compress256` is no longer public with no obvious replacement. The fact that the function is still marked `pub` implies that this isn't an intentional change.

We use the compression function over at https://github.com/noir-lang/noir so it would be great to be able to continue using the implementation from this crate going forwards.

TomAFrench:
Closes #732 

This PR replaces the exports for the sha2 compression functions that were removed in error.

### DIFF
--- a/sha2/src/lib.rs
+++ b/sha2/src/lib.rs
@@ -33,6 +33,9 @@ mod consts;
 mod sha256;
 mod sha512;
 
+pub use sha256::compress256;
+pub use sha512::compress512;
+
 digest::buffer_fixed!(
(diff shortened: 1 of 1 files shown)


TomAFrench:
Ah, my bad. This is accessible through `sha2::block_api::compress256`. Will leave this up to you on whether you want the compression function exported from the crate root or not. Feel free to close this.

PR #734, opened 2025-09-09
chore: apply `unreachable_pub` lint to crates

TomAFrench:
This PR rolls out the `unreachable_pub` lint over all the crates, fixing any issues raised. These are either items which have not been exported for quite some time or are very clearly internal implementation details.

I've used an inner attribute to activate the lint - my preference for this is to use the linting system in [cargo.toml](https://doc.rust-lang.org/cargo/reference/manifest.html#the-lints-section) as it makes it easiest to enforce consistency over multiple crates but I didn't want to cause a split with the existing config.


md5/src/block_api.rs
```diff
@@ -10,7 +10,8 @@ use digest::{
     typenum::{U16, U24, U64, Unsigned},
 };
 
-use crate::{compress::compress, consts};
+pub use crate::compress::compress;
+use crate::consts;
```
TomAFrench: Changed to export `md5::block_api::compress` here like in `sha2`. I'm also good removing this to avoid this PR affecting the interface.

newpavlov:
I don't think I agree with this lint. It's pretty common in my experience to define a `pub(crate)` module and `pub` items in it with the implicit assumption that items will also be `pub(crate)`. Same with `pub(crate) struct` and `pub fn` methods on it. There are also the sealed traits which use unreachability of `pub` items intentionally (I know people want to add a proper support for it, but the proposals do not have a clear timeline).

belt-hash/src/lib.rs
```diff
@@ -6,6 +6,7 @@
 )]
 #![cfg_attr(docsrs, feature(doc_auto_cfg))]
 #![warn(missing_docs)]
+#![warn(unreachable_pub)]
```
newpavlov: It's better to merge the `warn`s: `#![warn(missing_docs, unreachable_pub)]`.

newpavlov:
@tarcieri
WDYT? Should we be more strict with `pub`s? If yes, then we would need to gradually enable the lint in other repos as well.

tarcieri:
I'm a fan of using `pub(crate)` when you mean `pub(crate)`, since it will prevent accidental leakage, though sealed traits are a notable exception.

If it's just `warn` though, you can always do `allow` where you need exceptions

ripemd/src/c320.rs
```diff
@@ -142,7 +142,7 @@ macro_rules! process_block(
     });
 );
 
-pub fn compress(h: &mut [u32; DIGEST_BUF_LEN], data: &[u8; 64]) {
+pub(crate) fn compress(h: &mut [u32; DIGEST_BUF_LEN], data: &[u8; 64]) {
```
newpavlov: It's better to use either `pub(crate)` or `pub(super)` for both the function and the constants.

md5/src/block_api.rs
newpavlov: I think it's better to do it in a separate PR. Do not forget to add a changelog entry.

newpavlov:
Ok, let's move forward with it then.

@TomAFrench 
Could you create similar PRs later in other repos as well?

blake2/src/consts.rs
```diff
@@ -1,6 +1,6 @@
 #![allow(clippy::unreadable_literal)]
 
-pub static SIGMA: [[usize; 16]; 12] = [
+pub(super) static SIGMA: [[usize; 16]; 12] = [
```
newpavlov: In other crates you used `pub(crate)` for constants.

md5/src/block_api.rs
TomAFrench: done.

ripemd/src/c320.rs
TomAFrench: Agreed, I've changed the remaining uses of `super` to basically just cover the architecture-specific compression functions as these are the only items which should definitely never be imported elsewhere in the crate.

md5/src/compress.rs
```diff
@@ -1,12 +1,17 @@
...
-        pub use loongarch64_asm::compress;
+        use loongarch64_asm::compress as compress_inner;
     } else {
         mod soft;
-        pub use soft::compress;
+        use soft::compress as compress_inner;
     }
 }
+
+/// md5 compression function
```
TomAFrench:
```suggestion
/// MD5 compression function
```

TomAFrench:
Should be good to go now

newpavlov (approved):

newpavlov:
Thank you!

PR #736, opened 2025-09-13
fix(simd): correct variable naming from efgh to hgfe in SHA-256 imple…

hengyuss:
i think the variable name is incorrect, it make me confusing, 
 I've reviewed the code and believe this variable should be renamed from efghto hgfe.
 
image
image


newpavlov:
It looks you are right, this probably was a typo. Thank you for reviewing the code and fixing the issue!

PR #740, opened 2025-10-02
Migrate from doc_auto_cfg to doc_cfg

patrickelectric:
https://github.com/rust-lang/rust/pull/138907


newpavlov (approved):
The CI failure is fixed in #741.

PR #741, opened 2025-10-02
gost94: block_api: Fix typo

patrickelectric:


baloo (commented):
Those typos may be intentional to avoid a future reserved keyword or something.

EDIT: oh, this actually causes issues with CI
https://github.com/RustCrypto/hashes/actions/runs/18195287950/job/51799825623?pr=740

Either that or add an entry in `extend-words` (https://github.com/RustCrypto/hashes/blob/c73bd68ebbb79cfa1a09eb35d7268d4bee9f6cc2/.typos.toml#L6) I'll let someone else chime in.

newpavlov (approved):

PR #751, opened 2025-11-04
`bash-prg-hash`: Initial implementation

makavity:
1. I am not sure to assert in block_api
2. I am not sure in `new` and `new_with_empty_header` functions.


makavity:
I am also not sure, it should be implemented as prg-hash.
Because AEAD algorithm uses `start`, `squeeze`, `absorb` and `encrypt` functions. But it is correct to implement `encrypt` here and make `block_api` methods - pub?

bash-prg-hash/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+readme = "README.md"
+edition = "2024"
+rust-version = "1.85"
+documentation = "https://docs.rs/belt-hash-prg"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["belt", "stb", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+
+[dependencies]
+digest = "0.11.0-rc.3"
```
newpavlov: Update to 0.11.

bash-prg-hash/tests/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+        BA286D16912EBBACEC3D8143966107F6"
+    )
+);
+
+test_bash_prg_rand!(
+    bashprg2562_rand,
+    BashPrgHash2562,
+    hex!(
+        "AD07A8D61928296F4115F9E51AAA5FA986899BFDA8443F139D969600064EBCE2
+        D591F583FA27F6B0F7E73DA2B29AF382AC2374C04463B91A27F1C48FEE8AAB2C"
```
newpavlov: Write it like this:
```rust
hex!(
    "AD07..."
    "D591..."
)
```
Same for `bashprg1921_rand`.

bash-prg-hash/tests/mod.rs
```diff
@@ -0,0 +1,48 @@
...
+digest::new_test!(bashprg2562, BashPrgHash2562, xof_reset_test);
+
+macro_rules! test_bash_prg_rand {
+    ($name:ident, $hasher:ty, $expected:expr) => {
+        #[test]
+        fn $name() {
+            use bash_prg_hash::{HashLevel, SecurityLevel};
+            let mut h = <$hasher>::default();
+            digest::dev::feed_rand_16mib(&mut h);
+            let mut output = vec![0u8; <<$hasher as HashLevel>::Level as SecurityLevel>::LEVEL / 4];
```
newpavlov: I think `let mut output = [0u8; <<$hasher as HashLevel>::Level as SecurityLevel>::LEVEL / 4];` should work since `$hasher` is a concrete type here.

bash-prg-hash/src/oids.rs
```diff
@@ -0,0 +1,30 @@
+use digest::const_oid::{AssociatedOid, ObjectIdentifier};
+
+use crate::{
+    block_api::BashPrgHashCore,
+    variants::{Cap1, Cap2, Level128, Level192, Level256},
+};
+
+impl AssociatedOid for BashPrgHashCore<Level128, Cap1> {
```
newpavlov: We usually implement `AssociatedOid` for "buffered" types.

Also, use type aliases instead, i.e. `impl AssociatedOid for BashPrgHash1281 { ... }`?

bash-prg-hash/README.md
```diff
@@ -0,0 +1,63 @@
...
+let mut hasher = BashPrgHash2561::default();
+hasher.update(b"hello world!");
+
+let mut hash = [0u8; 32];
+hasher.finalize_xof_into(&mut hash);
+
+assert_eq!(hash, hex!("0C6B82907AE77386DDF0BA2D7CFDDD99F79A9B0094E545AEF8968A99440F5185"));
+
+// Hex-encode hash using https://docs.rs/base16ct
+let hex_hash = base16ct::upper::encode_string(&hash);
```
newpavlov: We removed such examples from other crates, so you can delete it and remove the `base16ct` dependency.

bash-prg-hash/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+### Contribution
+
+Unless you explicitly state otherwise, any contribution intentionally submitted
+for inclusion in the work by you, as defined in the Apache-2.0 license, shall be
+dual licensed as above, without any additional terms or conditions.
+
+[//]: # (badges)
+
+[crate-image]: https://img.shields.io/crates/v/belt-hash.svg
```
newpavlov: Wrong crate name here and below.

bash-prg-hash/README.md
```diff
@@ -0,0 +1,63 @@
...
+hasher.finalize_xof_into(&mut hash);
+
+assert_eq!(hash, hex!("0C6B82907AE77386DDF0BA2D7CFDDD99F79A9B0094E545AEF8968A99440F5185"));
+
+// Hex-encode hash using https://docs.rs/base16ct
+let hex_hash = base16ct::upper::encode_string(&hash);
+assert_eq!(hex_hash, "0C6B82907AE77386DDF0BA2D7CFDDD99F79A9B0094E545AEF8968A99440F5185");
+```
+
+Also, see the [examples section] in the RustCrypto/hashes readme.
```
newpavlov: The examples section was removed and instead now we refer users to the `digest` crate.

bash-prg-hash/LICENSE-MIT
```diff
@@ -0,0 +1,25 @@
+Copyright (c) 2025 The RustCrypto Project Developers
```
newpavlov: Bump to 2026.

bash-prg-hash/src/variants.rs
```diff
@@ -0,0 +1,72 @@
...
+
+/// Sealed trait to prevent external implementations.
+pub trait Sealed {}
+
+/// Security level trait for programmable algorithms.
+///
+/// Specified in Section 5.3 of STB 34.101.77-2020.
+///
+/// Standard levels: ℓ ∈ {128, 192, 256}.
+pub trait SecurityLevel: Sealed {
```
newpavlov: I think we can use `typenum` types directly without introducing new types. Something like this:
```
// Sealed traits
pub trait SecurityLevel {}
pub trait Capacity {}

impl SecurityLevel for U128 {}
impl SecurityLevel for U192 {}
impl SecurityLevel for U256 {}

impl Capacity for U1 {}
impl Capacity for U2 {}
```

And document accepted generic parameters in `BashPrgHash` docs.

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+
+/// bash-prg-hash hasher generic over security level and capacity.
+#[derive(Clone)]
+pub struct BashPrgHash<L: SecurityLevel, D: Capacity> {
+    core: BashPrgHashCore<L, D>,
+    finalized: bool,
+}
+
+/// Helper trait to extract security level from hash type
+pub trait HashLevel {
```
newpavlov: Maybe you should use [`CollisionResistance`](https://docs.rs/digest/latest/digest/trait.CollisionResistance.html) instead?

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+use digest::{ExtendableOutput, ExtendableOutputReset, Reset, Update, XofReader};
+
+pub use block_api::BashPrgHashCore;
+pub use variants::{Cap1, Cap2, Capacity, Level128, Level192, Level256, SecurityLevel};
+
+/// bash-prg-hash hasher generic over security level and capacity.
+#[derive(Clone)]
+pub struct BashPrgHash<L: SecurityLevel, D: Capacity> {
+    core: BashPrgHashCore<L, D>,
+    finalized: bool,
```
newpavlov: Why do you need `finalized` flag? It should be impossible to use "dirty" state? Users either consume it with `ExtendableOutput` or get reset state with `ExtendableOutputReset`.

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+impl<L: SecurityLevel, D: Capacity> XofReader for BashPrgHashReader<L, D> {
+    fn read(&mut self, buffer: &mut [u8]) {
+        self.core.squeeze(buffer);
+    }
+}
+
+impl<L: SecurityLevel, D: Capacity> Clone for BashPrgHashReader<L, D> {
+    fn clone(&self) -> Self {
```
newpavlov: This can be replaced with `#[derive(Clone)]` if `typenum` types are used for `SecurityLevel` and `Caapcity`.

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    type Level: SecurityLevel;
+}
+
+impl<L: SecurityLevel, D: Capacity> HashLevel for BashPrgHash<L, D> {
+    type Level = L;
+}
+
+impl<L: SecurityLevel, D: Capacity> BashPrgHash<L, D> {
+    /// Create a new hasher with an announcement (header).
+    pub fn new(header: &[u8]) -> Self {
```
newpavlov: I am not completely sure, but it looks like "announcement" plays role of "customization", no? If so, then you can implement [`CustomizedInit`](https://docs.rs/digest/latest/digest/trait.CustomizedInit.html) instead of using inherent methods.

One wrinkle is that the spec places restriction on "announcement"  length, so we probably need a fallible version of `CustomizedInit`.

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,260 @@
...
+/// Specified in Section 8.12 of STB 34.101.77-2020.
+pub struct BashPrgHashCore<L: SecurityLevel, D: Capacity> {
+    state: [u64; STATE_WORDS],
+    rate_bytes: usize,            // r/8 - buffer size in bytes
+    offset: usize,                // current offset in bytes
+    header: [u8; MAX_HEADER_LEN], // max header size (480 bits = 60 bytes)
+    header_len: usize,            // header length in bytes
+    data_committed: bool,         // whether commit(DATA) was called in
+    _level: PhantomData<L>,
+    _capacity: PhantomData<D>,
```
newpavlov: You can write it as `_pd: PhantomData<(L, D)>`.

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,260 @@
...
+const DATA: u8 = 0b000010;
+/// Data type codes from Table 3 of STB 34.101.77-2020
+const OUT: u8 = 0b000100;
+
+/// Core bash-prg-hash hasher state generic over security level and capacity.
+///
+/// Specified in Section 8.12 of STB 34.101.77-2020.
+pub struct BashPrgHashCore<L: SecurityLevel, D: Capacity> {
+    state: [u64; STATE_WORDS],
+    rate_bytes: usize,            // r/8 - buffer size in bytes
```
newpavlov: Field documentation is usually written like this:
```rust
/// Foo docs
foo: u32,
```

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,260 @@
...
+
+/// Data type codes from Table 3 of STB 34.101.77-2020
+const DATA: u8 = 0b000010;
+/// Data type codes from Table 3 of STB 34.101.77-2020
+const OUT: u8 = 0b000100;
+
+/// Core bash-prg-hash hasher state generic over security level and capacity.
+///
+/// Specified in Section 8.12 of STB 34.101.77-2020.
+pub struct BashPrgHashCore<L: SecurityLevel, D: Capacity> {
```
newpavlov: I wouldn't call it a "core" type since it absorbs byte slices and not blocks.

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,260 @@
...
+    state: [u64; STATE_WORDS],
+    rate_bytes: usize,            // r/8 - buffer size in bytes
+    offset: usize,                // current offset in bytes
+    header: [u8; MAX_HEADER_LEN], // max header size (480 bits = 60 bytes)
```
newpavlov: IIUC `header` is consumed first, so you don't need to keep it around. You can consume it immediately as part of initialization. It would mean that you will not be able to implement the reset traits, but I think it's fine.

newpavlov (commented):
Sorry for the late review!

Some preliminary comments without looking deep at the implementation and the spec.

makavity:
Thanks for review, @newpavlov.
As I see, all fixed

makavity:
Looks like the `sha1` is failed because timeout, not because PR changes.

bash-prg-hash/README.md
newpavlov: Sorry, I meant only the `base16ct` part, it's worth to keep the rest of the example.

bash-prg-hash/tests/mod.rs
newpavlov: I meant that you don't need `vec!` here and can use plain arrays instead.

bash-prg-hash/src/block_api.rs
newpavlov: You haven't fixed it, no?

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,241 @@
...
+/// Data type codes from Table 3 of STB 34.101.77-2020
+const OUT: u8 = 0b000100;
+
+/// bash-prg-hash hasher state generic over security level and capacity.
+///
+/// Specified in Section 8.12 of STB 34.101.77-2020.
+pub struct BashPrgHashState<L: SecurityLevel, D: Capacity> {
+    state: [u64; STATE_WORDS],
+    /// r/8 - buffer size in bytes
+    rate_bytes: usize,
```
newpavlov: You don't need to keep the rate as part of the state since it's  a constant computable from the generic parameters.

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,241 @@
...
+    /// current offset in bytes
+    offset: usize,
+    /// whether commit(DATA) was called
+    data_committed: bool,
```
newpavlov: This flag is not necessary.

bash-prg-hash/tests/mod.rs
makavity: Ah, you're right

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+///
+/// let mut hasher = BashPrgHash1281::default();
+/// hasher.update(b"hello world");
+/// let mut reader = hasher.finalize_xof();
+/// let mut output = [0u8; 32];
+/// reader.read(&mut output);
+/// ```
+#[derive(Clone)]
+pub struct BashPrgHash<L: SecurityLevel, D: Capacity> {
+    state: BashPrgHashState<L, D>,
```
newpavlov: Why have a separate `BashPrgHashState` type? As I wrote int the other comment, it's not a "core" type either way, so you can just inline its implementation and remove the `block_api` module completely.

bash-prg-hash/src/block_api.rs
makavity: image
Fixed, as I see

newpavlov: Huh, wtf... I see old version of the code then on the diff page for some reason. Looks like a GH issue again... Could you squash all commits and force push them?

bash-prg-hash/src/lib.rs
makavity: I think it is better not to inline, but separate type, smth like `Bash`. Because it is also used in AEAD bash.

bash-prg-hash/src/block_api.rs
makavity: Use offset as marker? For example, is commit needed == usize::MAX?

bash-prg-hash/tests/mod.rs
```diff
@@ -0,0 +1,50 @@
...
+            let mut h = <$hasher>::default();
+            digest::dev::feed_rand_16mib(&mut h);
+            let mut output = [0u8;
+                <<$hasher as CollisionResistance>::CollisionResistance as Unsigned>::USIZE * 2];
+            h.finalize_xof_into(&mut output);
+            assert_eq!(&output[..], $expected);
+        }
+    };
+}
+//
```
newpavlov: Empty comment.

bash-prg-hash/src/block_api.rs
newpavlov: You commit only as part of the finalization process, so why do you need to track whether you currently finalize or not as part of the state?

bash-prg-hash/src/lib.rs
newpavlov: But you don't have any functionality exposed for this type except of some basic trait impls.

You haven't submitted implementation of this AEAD mode yet, correct? It will depend on how much code duplication it will save us, it may be worth to just vendor specialized implementation similarly to what we do with Ascon.

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+/// use bash_prg_hash::{BashPrgHash1281, digest::{ExtendableOutput, Update, XofReader}};
+///
+/// let mut hasher = BashPrgHash1281::default();
+/// hasher.update(b"hello world");
+/// let mut reader = hasher.finalize_xof();
+/// let mut output = [0u8; 32];
+/// reader.read(&mut output);
+/// ```
+#[derive(Clone)]
+pub struct BashPrgHash<L: SecurityLevel, D: Capacity> {
```
newpavlov: It may be easier to use rate in bytes instead of capacity, e.g. see the `cshake` crate.

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,133 @@
...
+impl<L: SecurityLevel, D: Capacity> fmt::Debug for BashPrgHash<L, D> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("BashPrgHash { ... }")
+    }
+}
+
+impl<L: SecurityLevel, D: Capacity> digest::CollisionResistance for BashPrgHash<L, D>
+where
+    L: Div<digest::typenum::U8>,
+    <L as Div<digest::typenum::U8>>::Output: Unsigned,
```
newpavlov: To not deal with this annoying bounds you can just write 3 separate impls for each capacity/rate.

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,227 @@
...
+        }
+    }
+}
+
+impl<L: SecurityLevel, D: Capacity> Clone for BashPrgHashState<L, D> {
+    fn clone(&self) -> Self {
+        Self {
+            state: self.state,
+            offset: self.offset,
+            _pd: PhantomData,
```
newpavlov: You can use `#[derive(Clone)]`

bash-prg-hash/src/block_api.rs
```diff
@@ -0,0 +1,227 @@
...
+/// Data type codes from Table 3 of STB 34.101.77-2020
+const OUT: u8 = 0b000100;
+
+/// bash-prg-hash hasher state generic over security level and capacity.
+///
+/// Specified in Section 8.12 of STB 34.101.77-2020.
+pub struct BashPrgHashState<L: SecurityLevel, D: Capacity> {
+    state: [u64; STATE_WORDS],
+    /// current offset in bytes
+    offset: usize,
```
newpavlov: I think `u8` should be sufficient here?

makavity: Is that ok with a lot of casts to usize?

newpavlov: You can use `From`/`Into` to transform `u8` into `usize`. For the other direction you would need to use `TryFrom/Into` with `expect`.

I would like to tweak the absorption code a bit overall. I will try to draft an example using `cshake` later.

bash-prg-hash/src/lib.rs
makavity: Don't think so, because anyway we need two generic params, no?

makavity: I think it is not possible here, because use `U128` as `rate` is not possible:
```
1.	ℓ=128, d=1: r = (1536 - 21128)/8 = (1536 - 256)/8 = 1280/8 = 160 bytes → U160 	
2.	ℓ=128, d=2: r = (1536 - 22128)/8 = (1536 - 512)/8 = 1024/8 = 128 bytes → U128 	
3.	ℓ=192, d=1: r = (1536 - 21192)/8 = (1536 - 384)/8 = 1152/8 = 144 bytes → U144 	
4.	ℓ=192, d=2: r = (1536 - 22192)/8 = (1536 - 768)/8 = 768/8 = 96 bytes → U96 	
5.	ℓ=256, d=1: r = (1536 - 21256)/8 = (1536 - 512)/8 = 1024/8 = 128 bytes → U128 <- conflicts with ℓ=128, d=2
6.	ℓ=256, d=2: r = (1536 - 22256)/8 = (1536 - 1024)/8 = 512/8 = 64 bytes → U64
```

newpavlov: Ah, true, I forgot about the multiplication on capacity (IMO this parameter is somewhat weird, but oh well...). In that case you can use security level in bytes. It would allow you to remove division in `CollisionResistance` impls and simplify the rate computation to `192 - L::USIZE*D::USIZE`.

bash-prg-hash/src/block_api.rs
makavity: Did it, but i think, use offset in usize is more cleaner

bash-prg-hash/src/lib.rs
makavity: Almost.
`192 - 2 * L::USIZE*D::USIZE`

newpavlov:
I added `TryCustomizedInit` trait in digest v0.11.3. Also, take a look at #849 (the change was inspired by this PR), it should help to simplify the implementation.

makavity:
> I added `TryCustomizedInit` trait in digest v0.11.3. Also, take a look at #849 (the change was inspired by this PR), it should help to simplify the implementation.

Cannot implement squeeze.
In `SpongeCursor::squeeze_u64_le`:
```rust
for block in &mut blocks {
    sponge(state);

    let mut dst_chunks = block.chunks_exact_mut(size_of::<u64>());
    for (src, dst_chunk) in state.iter_mut().zip(&mut dst_chunks) {
        dst_chunk.copy_from_slice(&src.to_le_bytes());
    }
    assert!(dst_chunks.into_remainder().is_empty());
}
```
This does permute then read, while `bash-prg-hash` needs read then permute for the first block after `commit(OUT)` (and then permutation between full-rate blocks).
So the first output block must come from the current committed state, without an extra permutation. With current API, that path cannot be expressed directly, so reader `squeeze` still needs custom logic.

Should I left it as-is, or maybe another mode in `SpongeCursor`?

makavity:
I've commited it without `squeeze` from `SpongeCursor`, have no idea how to use it in current approach.

bash-prg-hash/src/block_api.rs
makavity: Moved to `SpongeCursor` (partially)

newpavlov:
>Should I left it as-as, or maybe another mode in SpongeCursor?

Can not say right now, I will need to read the spec first.

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,295 @@
...
+            state: hasher_clone.state,
+            cursor: hasher_clone.cursor.clone(),
+        }
+    }
+}
+
+impl<const RATE: usize, const CAPACITY: usize> Reset for BashPrgHash<RATE, CAPACITY> {
+    #[inline]
+    fn reset(&mut self) {
+        *self = Self::default();
```
newpavlov: This is not correct for customized states. We either should remove implementations of `Reset` traits or introduce separate "customizable" types. I think the former will be the easiest, though it would mean that we would not be able to use `xof_reset_test` for testing.

makavity: There is `restart` command in specification. Maybe we should implement it like that?
image


Sorry for the russian specification. There is no english version yet. RFC is in the draft now.

newpavlov: IIUC this command is for implementing the AEAD mode, no? It's not relevant to the discussed `Reset` trait. Its implementation should "reset" the hasher state to its initial initialized state, while here we reset potentially customized hashers to the non-customized state.

makavity: This command can be used in hashing mode. It is the "Reset to initial state".
Look for the: If `K != _|_`  - there is "if no key provided", thats for hash.

makavity: > It's not relevant to the discussed Reset trait. It's implementation should "reset" the hasher state to its initial initialized state, while here we reset potentially customized hashers to the non-customized state.

See your point. If init was with header (`try_new_customized`) it will be reset.

makavity: For the `Reset` trait we can store header in struct. Would it be correct?

newpavlov: But then you have to provide the "announcement" value, which we no longer have.

If you want to keep the command, then it's better to introduce a separate customizable type (or generic parameter) which would keep the customization string in its state as `[u8; 61]` (60 byte capacity + 1 byte for length). I am not sure it's worth the trouble.

makavity: You're right. I will remove it.

bash-prg-hash/tests/mod.rs
```diff
@@ -0,0 +1,50 @@
+use bash_prg_hash::{BashPrgHash1282, BashPrgHash1921, BashPrgHash2562};
+use digest::ExtendableOutput;
+use digest::dev::xof_reset_test;
+use hex_literal::hex;
+
+// Test vectors from STB 34.101.77-2020 (Appendix A, Table A.5)
+digest::new_test!(bashprg1282, BashPrgHash1282, xof_reset_test);
+digest::new_test!(bashprg1921, BashPrgHash1921, xof_reset_test);
+// Not in STB 34.101.77-2020, but included for completeness
+digest::new_test!(bashprg2562, BashPrgHash2562, xof_reset_test);
```
newpavlov: Are there any vectors for customized hashing? IIUC they are not part of the STB. It may be worth to generate synthetic tests for it.

bash-prg-hash/src/lib.rs
newpavlov: I already have changed it locally, so I will commit it shortly.

bash-prg-hash/tests/mod.rs
makavity: I'll check the testing methodic and will extend the tests.

makavity: Found it in the methodics.
https://apmi.bsu.by/assets/files/std/met-v10.zip

I'll add it and tweak xof_test for the `try_new_customized`.

newpavlov: See the `cshake` test for an example. We probably will add a function for this into a future version of `digest`.

newpavlov:
I refactored the code a fair bit. Feel free to comment on the changes if something is not clear or if you have suggestions on how we could improve it.

I think this PR is mostly ready for merge. We only need to release `sponge-cursor` and add tests for customized hashes.

bash-prg-hash/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+        let word_pos = pos / 8;
+        let byte_pos = pos % 8;
+
+        // Step 1: S[pos...pos+8) <- S[pos...pos+8) ⊕ (t || 01).
+        const OUT_PAD: u8 = (OUT << 2) | 0x01;
+        self.state[word_pos] ^= u64::from(OUT_PAD) << (8 * byte_pos);
+
+        // Step 2: S[r] <- S[r] ⊕ 1, where r = 1536 - 2 d ℓ (bit index).
+        const { assert!(RATE % 8 == 0) }
+        self.state[RATE / 8] ^= 1u64 << 7;
```
newpavlov, line 157: This looks a bit weird, but I guess it's a consequence of mixing little-endian byte order and big-endian bit order in the spec and `commit` acting on the state outside of the rate part. For comparison, in `cshake` and `turboshake` we use `self.state[RATE / 8 - 1] ^= 1 << 63;`.

makavity:
> I refactored the code a fair bit. Feel free to comment on the changes if something is not clear or if you have suggestions on how we could improve it.

Looks like the commit `1dc21053` broke the customization flow.
I'll commit the tests with announce, which worked before commit and continue to investigate.

makavity:
Just misspell in the header length calculation.

makavity:
@newpavlov in the methodic, there are test with 1000000 of `0x61` for input.
Should we add that tests for the completeness, or we can just skip it?

makavity:
> I refactored the code a fair bit. Feel free to comment on the changes if something is not clear or if you have suggestions on how we could improve it.

Really a fair bit. Now it's just two little functions :D
Awesome work, thank you!

bash-prg-hash/src/lib.rs
makavity: You're right, I suppose. I've noticed earlier, that spec is using first byte after rate.

newpavlov:
>there are test with 1000000 of 0x61 for input.

I think it's fine to add it (e.g. by feeding 1k byte chunks 1k times). It should be relatively fast.

bash-prg-hash/README.md
newpavlov: It looks like fix of this part was accidentally reverted.

newpavlov:
Please merge master (or rebase) and fix the [readme issue](https://github.com/RustCrypto/hashes/pull/751#discussion_r3196471314).

bash-prg-hash/Cargo.toml
```diff
@@ -0,0 +1,31 @@
...
+readme = "README.md"
+edition = "2024"
+rust-version = "1.85"
+documentation = "https://docs.rs/bash-prg-hash"
+repository = "https://github.com/RustCrypto/hashes"
+keywords = ["belt", "stb", "hash", "digest"]
+categories = ["cryptography", "no-std"]
+
+[dependencies]
+digest = "0.11"
```
newpavlov: Use `default-features = false` (see the other sponge-based crates).

PR #754, opened 2025-11-05
blake2: restore `Blake2bVar`/`Blake2sVar`

tarcieri:
Restores hashes that previously impl'd the now removed `VariableOutput` trait which were removed in #744. They're used by `argon2`.

Note the API has been transition from traits to inherent methods, and the implementations have not been restored for other runtime-variable hash functions, just `blake2` due to the need to use `Blake2bVar` in the `argon2` crate.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -72,9 +72,9 @@ dependencies = [
 
 [[package]]
 name = "blobby"
-version = "0.4.0"
+version = "0.4.0-pre.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "89af0b093cc13baa4e51e64e65ec2422f7e73aea0e612e5ad3872986671622f1"
+checksum = "e4db6eec520c1f3d3b060c5b67527c3d1d4122d9f0ff47edf0149a7ffa34cebb"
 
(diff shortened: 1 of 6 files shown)


newpavlov:
I think we should discourage use of runtime-variable output sizes. In the case of `argon2`, I think we could use the core type directly. It may be even a better fit, considering that we need the variable output only for the very last iteration of the loop.

tarcieri:
@newpavlov in the interim I need a stopgap as this is blocking the whole `rand_core` upgrade, and by even releasing `blake2` I broke the `argon2` prereleases: https://github.com/RustCrypto/password-hashes/issues/736

If you can modify `argon2` to avoid using this API, we can remove this again, but for now we need it.

If you'd like, I can slap a `#[deprecated]` on it for now.

newpavlov:
Give me ~30 minutes. I will try to draft a fix which does not rely on this PR. If I am not finished by then, you can merge this PR.

tarcieri:
@newpavlov wrong PR, I assume

tarcieri:
These usages were removed in https://github.com/RustCrypto/password-hashes/pull/739

PR #755, opened 2025-11-06
md5: minor optimization in software backend

newpavlov:
Replaces the logical OR in the G function with addition. It seemingly results in a better ALU utilization and improves performance by several percents. From 699 MB/s to 753 MB/s on my x86 PC and from 910 MB/s to 960 MB/s on Mac M4.

Based on #749

### DIFF
--- a/md5/src/compress/soft.rs
+++ b/md5/src/compress/soft.rs
@@ -12,10 +12,10 @@ fn op_f(w: u32, x: u32, y: u32, z: u32, m: u32, c: u32, s: u32) -> u32 {
 }
 #[inline(always)]
 fn op_g(w: u32, x: u32, y: u32, z: u32, m: u32, c: u32, s: u32) -> u32 {
-    // Here we replace the logical OR in `(x & z) | (y & !z)` with addition.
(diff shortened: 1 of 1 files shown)


newpavlov:
Huh... Compiling `op_g` in isolation results in the same assembly whether we use `wrapping_add` or not (see [here](https://rust.godbolt.org/z/Y58z99Pza)). In other words, the compiler is able to apply such optimization itself. But it seems this change (accidentally?) nudges the compiler towards a better codegen.

newpavlov:
Interestingly, compiling the full compress function with the `wrapping_add` change results in [more instructions](https://rust.godbolt.org/z/azfa3oMqT) on x86 (672 vs 648), but I guess the resulting code is a bit friendlier to pipeline. On AArch64 the number of instructions is the same (632), but with `wrapping_add` some `orr` instructions get replaced with `add` (i.e. the compiler has failed to apply the optimization when the full function is compiled).

@tarcieri 
Could you check whether this change results in a better performance on Mac?

newpavlov:
On M4 this change improves performance from 910 MB/s to 960 MB/s.

tarcieri:
Seems about 5-6% faster on my M1 Max

PR #757, opened 2025-11-15
fix: clarify operator precedence in key validation

Lil-Duckling-22:
Add explicit parentheses to key validation condition to make operator precedence clear. The logic remains unchanged, but the grouping is now explicit and easier to understand during code review.


newpavlov:
Closing as an automated PR.

PR #759, opened 2025-11-25
streebog: fix bit-length counter carry propagation to 512-bit

radik878:
The update_n() implementation propagated carry only through limbs 1..6, effectively making n a 448-bit counter. According to GOST R 34.11-2012 (RFC 6986), the total processed bit-length must be tracked modulo 2^512. This change extends the loop to include limb 7 so carry ripples through all eight 64-bit words and any further carry is discarded, thus restoring correct modulo 2^512 semantics. Although this bug only manifests for astronomically large inputs, it is a correctness issue and brings the implementation in line with the specification and the behavior of similar counters in this codebase.


newpavlov (approved):
Huh, I wonder why `1..7` was used initially.

Thank you for noticing this!

PR #761, opened 2025-12-01
perf(k12): remove redundant 8KiB buffer zeroing in process_chunk

Galoretka:
The buffer was zeroed after each full chunk in KangarooTwelveCore::process_chunk(). This is unnecessary because subsequent reads only consume [..bufpos] and bufpos is reset to 0. Removing the memset avoids extra work on long inputs and aligns with the project’s pattern of cleaning sensitive data via Drop under the zeroize feature.

### DIFF
--- a/k12/src/block_api.rs
+++ b/k12/src/block_api.rs
@@ -47,7 +47,6 @@ impl<'cs> KangarooTwelveCore<'cs> {
         }
 
         self.chain_length += 1;
-        self.buffer = [0u8; CHUNK_SIZE];
         self.bufpos = 0;
     }
 
--- a/k12/src/block_api.rs
+++ b/k12/src/block_api.rs
@@ -39,7 +39,7 @@ impl<'cs> KangarooTwelveCore<'cs> {
     }
 
     fn process_chunk(&mut self) {
-        debug_assert!(self.bufpos == CHUNK_SIZE);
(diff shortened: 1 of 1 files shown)


tarcieri:
> This is unnecessary because subsequent reads only consume [..bufpos]

I would feel better about that if there were a `debug_assert_eq!(self.bufpos, CHUNK_SIZE)` for the `if self.chain_length == 0` above, though I guess if that weren't the case it could be unnecessarily hashing zeros

newpavlov:
I wonder why the implementation simply does not use `CHUNK_SIZE` for block size.

tarcieri:
@newpavlov IIUC, the block size is implicitly `CHUNK_SIZE` unless chaining is being used

Galoretka:
> > This is unnecessary because subsequent reads only consume [..bufpos]
> 
> I would feel better about that if there were a `debug_assert_eq!(self.bufpos, CHUNK_SIZE)` for the `if self.chain_length == 0` above, though I guess if that weren't the case it could be unnecessarily hashing zeros

corrected

tarcieri:
Oof, sorry this one went by the wayside.

@Galoretka can you rebase?

newpavlov:
I am working on a completely new implementation of `k12` which does not use 8 KiB buffers (it absorbs partial chunk data immediately), so I think we can close this PR.

PR #767, opened 2026-01-03
Rename `ascon-hash` => `ascon-hash256`

tarcieri:
We no longer control the `ascon-hash` name.

This follows a similar rename we did for AEADs: `ascon-aead` =>  `ascon-aead128`


ascon-hash256/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
-name = "ascon-hash"
-version = "0.3.0-pre"
+name = "ascon-hash256"
+version = "0.5.0-pre"
```
tarcieri, line 3: Note: picked to match `ascon` v0.5

PR #785, opened 2026-02-13
Remove README examples

newpavlov:
The examples now reside in the `digest`'s README/docs. This PR also updates references in the implementation crate READMEs and removes `base16ct` examples from them. I plan to add `base16ct` examples to `digest` in a later PR.


tarcieri (approved):
Huh, didn't realize every crate was documenting hex serialization, and those toplevel examples were untested? Seems bad.

PR #790, opened 2026-02-17
sha3: use `KeccakP1600` struct for state

tarcieri:
Added in RustCrypto/sponges#107

This struct can automatially make use of CPU intrinsics when they are available. Currently only supports ARMv8's `FEAT_SHA3` intrinsics.


newpavlov (approved):
I want to try the suggested callback-based API, but it can be done in separate PRs.

PR #791, opened 2026-02-18
sha2: Add tests that check stability of the serialized state

wiktor-k:
Hi,

As mentioned in other tickets (e.g. https://github.com/RustCrypto/traits/issues/1693) the serializable state format should be stable across minor version releases. As we're planning to use it in one project I think it would be nice if there were tests which capture this property.

These tests use SHA-512 and serialize its state, comparing it to a well-known values. Then, they rebuild the hasher from that state and check if the final digest is correctly calculated.

As documented in the compatibility section fro the `SerializableState` these values should be stable for semver-compatible releases.

See: https://github.com/RustCrypto/traits/issues/1693

If there is a better approach I'm open for suggestions, filing this PR as a way of starting the discussion. I hope you don't mind :bow:


newpavlov:
We already have such tests, see: https://github.com/RustCrypto/hashes/blob/master/sha2/tests/mod.rs#L15-L20

wiktor-k:
Oh, indeed, that's reassuring! I've migrated from `0.11.0-rc.0` to `0.11.0-rc.5`, noticed changes and added the test before checking the existing ones. Sorry for the noise! :pray:

PR #796, opened 2026-03-06
Migrate to closure-based `keccak`

newpavlov:


sha3/src/block_api.rs
```diff
@@ -271,7 +279,8 @@ where
         for (src, dst) in self.state.iter().zip(block.chunks_mut(8)) {
             dst.copy_from_slice(&src.to_le_bytes()[..dst.len()]);
         }
-        keccak::p1600(&mut self.state, ROUNDS);
+        self.keccak
+            .with_p1600::<ROUNDS>(|p1600| p1600(&mut self.state));
```
newpavlov, line 278: Unfortunately, `XofReaderCore` has only one `read_block` method, so we have to call `with_p1600` (which contains potential branching) on each generated block. In a future version of `digest` it may be worth to introduce additional methods to amend this.

PR #798, opened 2026-03-13
sha3: Add docs to clarify the `DS` generic param on TurboShake

rozbb:


sha3/src/turbo_shake.rs
```diff
@@ -16,7 +16,8 @@ macro_rules! impl_turbo_shake {
         $name:ident, $reader_name:ident, $rate:ty, $alg_name:expr
     ) => {
         #[doc = $alg_name]
-        #[doc = " hasher."]
+        #[doc = " hasher with domain separator `DS`. `DS` MUST be in the range `0x01..=0x7f`. \
+        Default is `0x1f`."]
```
newpavlov: It's probably worth to use `const DS: u8 = 0x1F` below.

Also, please tweak the doc, so result would look like this:
```rust
/// $alg_name hasher with domain separator `DS`.
///
/// `DS` MUST be in the range `0x01..=0x7f`, the default value is `0x1F`.
/// Using `DS` outside of the range will result in a panic during initialization.
```

PR #802, opened 2026-03-20
sha2: refactor backends

newpavlov:
This PR reorganizes backends structure and related configuration flags. Now backends like `x86_sha` expose `compress` function marked with `#[target_feature(enable = "...")]` which has to be called using `unsafe` after an appropriate run-time or compile-time check. Autodetection is now handled by the (default) fallback `cfg_if!` branch instead of doing it in the backend modules. `sha2_backend` now forces selection of the specified backend and if necessary checks at compile time whether target features required by the backend were properly enabled.

`sha2_backend="soft-compact"` and `sha2_backend="riscv-zknh-compact"` configuration flags are replaced by `sha2_backend_soft="compact"` and `sha2_backend_riscv_zknh="compact"` flags respectively.

Additionally `sha2_256_backend` and `sha2_512_backend` configuration flags are introduced for finer-grained control over selected backends.


sha2/src/sha512.rs
```diff
@@ -2,31 +2,48 @@ cfg_if::cfg_if! {
...
-    ))] {
-        mod riscv_zknh_compact;
-        mod riscv_zknh_utils;
-        use riscv_zknh_compact::compress;
-    } else if #[cfg(target_arch = "aarch64")] {
-        mod soft;
-        mod aarch64_sha2;
-        use aarch64_sha2::compress;
+        target_arch = "aarch64",
+        sha2_backend = "aarch64-sha3",
```
newpavlov: Maybe we should introduce separate flags for SHA-256 and SHA-512 backends, i.e. `sha2_sha256_backend` and `sha2_sha512_backend`?

tarcieri (approved):

PR #804, opened 2026-03-21
SHA1 - NIST FIPS-180 2 NSRL Test Data

WingZer0o:
https://www.nist.gov/itl/ssd/software-quality-group/nsrl-test-data

I was doing some more reading on the topic of conversation for FIPS and wanted to try and implement some more code. I propose that we add 3 tests that are specified in the above document for SHA1.

- A file containing the ASCII string "abc" results in a 160 bit message digest of a9993e36 4706816a ba3e2571 7850c26c 9cd0d89d.
- A file containing the ASCII string "abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq"results in a 160-bit message digest of 84983e44 1c3bd26e baae4aa1 f95129e5 e54670f1.
- A file containing the binary-coded form of the ASCII string which consists of 1,000,000 repetitions of the character "a" results in a SHA-1 message digest of 34aa973c d4c4daa4 f61eeb2b dbad2731 6534016f.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -290,6 +290,7 @@ dependencies = [
  "cfg-if",
  "cpufeatures",
  "digest",
+ "hex",
  "hex-literal",
 ]
 
--- a/sha1/Cargo.toml
+++ b/sha1/Cargo.toml
@@ -21,6 +21,7 @@ cpufeatures = "0.3"
 
 [dev-dependencies]
 digest = { version = "0.11", features = ["dev"] }
+hex = "0.4.3"
 hex-literal = "1"
 
 [features]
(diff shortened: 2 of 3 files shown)


newpavlov:
We already test the "abc" and "abcdbcdecdefdefgefghfghighijhijkijkljklmklmnlmnomnopnopq" vectors as part of the `sha1_kat` test. I think we previously had the 1 million "a" test, but intentionally replaced it with `sha1_rand`.

Finally, use of the `hex` crate and `std` items is unnecessary.

WingZer0o:
@newpavlov Thanks for the feedback! I'll look into it a little more.

PR #815, opened 2026-04-01
sha3: move cSHAKE and TurboSHAKE into separate crates

newpavlov:


newpavlov:
@itzmeanjan
Would be you open to transferring ownership of the `turboshake` crate to us?

tarcieri (approved):

PR #831, opened 2026-04-14
turbo-shake: fix badges in readme

newpavlov:
Also slightly tweaks the crate description.


tarcieri:
What's happening with renaming it to `turboshake`?

newpavlov:
The owner has agreed to transfer the crate, but did not do it yet. Probably, he wants to review the code as I suggested. I will ping him in the transfer request issue.

PR #834, opened 2026-04-18
cshake: fix zero padding when data already aligns to block boundary

rpopplewell:
When the buffer already aligns with the block boundary, the existing implementation adds an additional block of zeros to the input. This occurs when the total length of the function name and customization string is a multiple of the block size.

This isn't compliant with the "bytepad" algorithm in section 2.3.3 of [SP 800-185](https://nvlpubs.nist.gov/nistpubs/SpecialPublications/NIST.SP.800-185.pdf). If the buffer already aligns with the block boundary, skip the padding.


newpavlov:
Thank you for reporting this!

It seems we had this bug in the original submission as well (#355), so we should probably also create a backport for sha3 v0.10. It's a bit surprising that this corner case is not covered by KATs. Could you add a regression test for this?

I also think that a slightly better fix will be to use the [`LazyBuffer`](https://docs.rs/block-buffer/latest/block_buffer/type.LazyBuffer.html) here instead of the branch.

rpopplewell:
Now using LazyBuffer and tests that fail on the old implementation. I [verified](https://gist.github.com/rpopplewell/e5999e1a693eed23e0a39d813c67ae5f) the outputs against the golang standard library.

newpavlov (approved):

PR #841, opened 2026-04-22
ascon-hash256: refactor code and split XOF into `ascon-xof128` crate

newpavlov:
Additionally, implements Ascon-CXOF128 as part of the new `ascon-xof128` crate.


ascon-xof128/src/block_api/cxof.rs
```diff
@@ -0,0 +1,149 @@
...
+    fn default() -> Self {
+        Self::new_customized(&[])
+    }
+}
+
+impl CustomizedInit for AsconCxof128Core {
+    #[inline]
+    fn new_customized(customization: &[u8]) -> Self {
+        let bit_len = 8 * customization.len();
+        assert!(bit_len < MAX_CUSTOM_LEN);
```
newpavlov: I am not sure about this panic. The NIST states:
>The length of the customization string **shall** be at most 2048 bits (i.e., 256 bytes).

Where **shall** is defined as:
>Term used to express a requirement that needs to be fulfilled to claim conformance to this standard.

_Technically_, the algorithm is able to work with longer customization strings (up to 2^64 bits), so we could skip this check and document it as a user's responsibility.

PR #849, opened 2026-05-02
Use `sponge-cursor` to remove unnecessary buffering

newpavlov:
Removes unnecessary buffering from sponge-based constructions using the `sponge-cursor` crate (see https://github.com/RustCrypto/utils/pull/1477).


newpavlov:
@tarcieri 
I think we can migrate `sha3` as well, but it would involve breaking changes (removal of `block_api` and change of the serialization format), so it would have to be done in a new breaking release. What do you think, should we migrate it or leave as-is?

tarcieri:
I think it’s fine to migrate though at this point I think you’d have to bump minor version, which I’m ok with

RoDmitry:
How do I use `sha3` with `hmac` after the removal of `block_api` then? Am I missing something?

newpavlov:
You can use [`SimpleHmac`](https://docs.rs/hmac/latest/hmac/struct.SimpleHmac.html).

baloo:
This also affects the use of EagerHash in ecdsa:
https://docs.rs/ecdsa/0.17.0-rc.18/ecdsa/struct.SigningKey.html#method.sign_digest_recoverable

That can't be used anymore with sha3 anymore. This show up in https://github.com/RustCrypto/elliptic-curves/blob/a64c16cc3909d4489a2b0300886aa9f0a1691ba6/k256/src/ecdsa.rs#L243

EDIT:
 that requirements comes from https://github.com/rustcrypto/signatures/pull/1076 but I'm sure **on this line** it actually applies.
 
 EDIT2:
 this also limits the use of dsa::SigningKey::sign_prehash_rfc6979: https://github.com/RustCrypto/signatures/blob/699547f9c3593578f6d28f4e73a282ba526c8626/dsa/src/signing_key.rs#L95

PR #852, opened 2026-05-07
sha2: replace x86 SHA-NI macros with const-generic function

remix7531:
Replaces the `rounds4!` and `schedule_rounds4!` macros in `sha256/x86_sha.rs` with `#[inline(always)]` functions parameterized by `const I: usize`.

This matches the function-based style already used elsewhere in `sha2`. The closest precedent is `sha512/x86_avx2.rs`, where caller-mut helpers like `rounds_0_63_avx(state: &mut State, x: &mut [__m128i; 8], ms: &mut MsgSchedule)` are plain `#[inline(always)] unsafe fn`.

In the function form, `const I: usize` keeps the `K32X4[I]` lookup and the `0x0E` shuffle constant compile-time, and `&mut __m128i` parameters cover the caller-binding mutation the macros were doing positionally.

Codegen is byte-identical to the macro version. Diffing `cargo rustc --release --emit=asm` on the `compress` symbol returns zero on both `-C target-cpu=x86-64` and `-C target-cpu=native`. All `sha2` tests pass.

The same pattern is used in `sha1/src/compress/x86_sha.rs`. Happy to do that as a follow-up if this lands.

### DIFF
--- a/sha2/src/sha256/x86_sha.rs
+++ b/sha2/src/sha256/x86_sha.rs
@@ -18,26 +18,28 @@ unsafe fn schedule(v0: __m128i, v1: __m128i, v2: __m128i, v3: __m128i) -> __m128
     _mm_sha256msg2_epu32(t3, v3)
 }
 
-macro_rules! rounds4 {
-    ($abef:ident, $cdgh:ident, $rest:expr, $i:expr) => {{
-        let k = crate::consts::K32X4[$i];
-        let kv = _mm_set_epi32(k[0] as i32, k[1] as i32, k[2] as i32, k[3] as i32);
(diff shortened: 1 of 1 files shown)


tarcieri (approved):

newpavlov:
Thank you! But I don't think we need const generics here, since the round number is used only for loading round constants (`std` uses const generics in intrinsics for immediate arguments, i.e. to effectively switch between different instructions which share one name).

I will close this PR in favor of #853. Could you do something similar for `sha1`?

remix7531:
Will do, const generic was not necessary. Having `i: usize` also produced the same asm to the version you justed merged.

PR #853, opened 2026-05-07
sha2: remove macros from `sha256/x86_sha`

newpavlov:
Additionally, performs a minor refactoring.

### DIFF
--- a/sha2/src/sha256.rs
+++ b/sha2/src/sha256.rs
@@ -20,11 +20,9 @@ cfg_if::cfg_if! {
 
         #[cfg(not(all(
             target_feature = "sha",
-            target_feature = "sse2",
-            target_feature = "ssse3",
             target_feature = "sse4.1",
         )))]
-        compile_error!("x86-sha backend requires sha, sse2, ssse3, sse4.1 target features");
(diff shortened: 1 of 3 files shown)


tarcieri (approved):

newpavlov:
With enabled optimizations the new code compiles down to the exactly same assembly ([old](https://rust.godbolt.org/z/MzYv8cax3) vs [new](https://rust.godbolt.org/z/9eWs1xve8)).

sha2/src/sha256.rs
```diff
@@ -20,11 +20,9 @@ cfg_if::cfg_if! {
 
         #[cfg(not(all(
             target_feature = "sha",
-            target_feature = "sse2",
-            target_feature = "ssse3",
             target_feature = "sse4.1",
         )))]
-        compile_error!("x86-sha backend requires sha, sse2, ssse3, sse4.1 target features");
+        compile_error!("x86-sha backend requires sha and sse4.1 target features");
```
newpavlov, line 25: Note that `sse4.1` implies `ssse3`, which implies `sse3`, which in turn (together with `sha`) implies `sse2`.

Issue #669, opened 2025-03-20
Problems using the md5 crate!


stela2502:
[dependencies]
md5 = "0.7.0"

in main.rs

use md5::{Md5, Digest};

And my cargo build throws this error:

 --> src\main.rs:2:11
  |
2 | use md5::{Md5, Digest};
  |           ^^^ no `Md5` in the root

I really do not understand what is going on! That should be used like that - yes?

I have updated my rust with rustup update and also cleaned out all build artifacts - I even tried it on Windows 11 and Ubuntu. I need your help. The epically simple code is here: https://github.com/stela2502/rust-geo-prep

THANK YOU FOR YOUR HELP!

newpavlov:
Our crate is called `md-5`, not `md5`.

sivizius:
This is on the front page of the [documentation](https://docs.rs/md-5/0.11.0/md5/):

1. `crate md5` as title

2. This example:

```rust
use md5::{Md5, Digest};
use hex_literal::hex;

let mut hasher = Md5::new();
hasher.update(b"hello world");
let hash = hasher.finalize();

assert_eq!(hash, hex!("5eb63bbbe01eeed093cb22bb8f5acdc3"));
```

I suggest to add some description to avoid any further confusion. I myself used the `md5` crate for a while and was somewhat annoyed by the incompatibility before eventually figuring out I had the wrong crate. Even though it isn’t the classic namesquatting (`md5` is actually 2 years older than `md-5` and the older one is afaics perfectly fine), it is somewhat confusing.

newpavlov:
We have this [note](https://github.com/RustCrypto/hashes#crate-names), but I guess it's worth to add something similar to the crate docs as well.

PR #867, opened 2026-05-12
md5: add note about library renaming

newpavlov:
The library renaming can be confusing for users, see: https://github.com/RustCrypto/hashes/issues/669


newpavlov:
Maybe it's worth to remove the renaming in a future breaking release?

tarcieri (approved):

PR #877, opened 2026-06-01
sha2: cross compile to uefi

baloo:
This is an attempt at reproducing an LLVM error I've seen in the field:

```
rustc-LLVM ERROR: Do not know how to split the result of this operator!
```

I have no clue where this actually comes from.

### DIFF
--- a/.github/workflows/sha2.yml
+++ b/.github/workflows/sha2.yml
@@ -152,6 +152,7 @@ jobs:
         target:
           - aarch64-unknown-linux-gnu
           - powerpc-unknown-linux-gnu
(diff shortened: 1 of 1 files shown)


baloo:
That doesn't reproduce in this repo, I've made a reproducer for the bug I'm chasing:
https://github.com/baloo/sha2-reproducer

I have no clue how to proceed for there.

newpavlov:
It's a known issue. See: https://github.com/rust-lang/rust/issues/117938

The recommended workaround is to use the `sha2_backend="soft"` configuration flag.

PR #886, opened 2026-06-24
Fix bash-hash package metadata and README links

daixihegu:
Updates bash-hash metadata and README badge links that were still pointing to belt-hash. This fixes the published documentation URL, crate/docs badges, workflow badge, and crate keyword so they consistently reference bash-hash.


newpavlov (approved):
Thank you!

PR #889, opened 2026-06-28
jh: implement SerializableState

Ap4sh:
part of #546

adds `SerializableState` for `JhCore` and exposes it through the public `Jh224`, `Jh256`, `Jh384`, and `Jh512` wrappers via `FixedHashTraits`

the serialized core state stores the 128-byte compressor state plus the processed block count, matching the wrapper-level serialization tests used by the other hash crates

tests:
- `cargo test -p jh --all-features`
- `cargo hack test --feature-powerset`
- `RUSTFLAGS='-Dwarnings' cargo +1.85.1 hack test --feature-powerset`
- `RUSTFLAGS='-Dwarnings' cargo +1.85.1 hack build --target thumbv7em-none-eabi --each-feature --exclude-features default,std`
- `RUSTFLAGS='-Dwarnings' cargo +1.85.1 hack build --target wasm32-unknown-unknown --each-feature --exclude-features default,std`
- `cargo fmt --all -- --check`
- `cargo clippy --all -- -D warnings`
- `typos`


newpavlov (approved):
I would prefer to resolve #546 as well since it's unclear on the first glance that it's correct to use `Compressor` finalization like this, but I guess it's fine to merge it now.

PR #900, opened 2026-07-11
jh: remove ppv-lite86 dependency

### LINKED ISSUE #546: jh: remove dependency on `ppv-lite86`

Ap4sh:
## Summary

- replace the `ppv-lite86` compressor with a portable `u128` implementation and an SSE2 backend for x86_64
- store the compressor as typed words so `zeroize` can wipe it without the previous unsafe byte cast
- write final and serialized state directly into destination buffers and check portable/SSE2 equivalence

SSE2 is part of the x86_64 baseline, so this keeps the accelerated path without runtime feature detection

Closes #546

## Testing

- `cargo +1.85.1 hack test --feature-powerset`
- `cargo +stable hack test --feature-powerset`
- `cargo +1.85.1 clippy --all -- -D warnings`
- JH cross-build matrix for `thumbv7em-none-eabi` and `wasm32-unknown-unknown` on MSRV and stable
- minimal versions check with the CI command
- `cargo +stable package -p jh --allow-dirty`

The 10,000-byte JH-256 bench stayed within noise at 292 MB/s before and 297 MB/s after

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -176,7 +176,6 @@ version = "0.2.0"
 dependencies = [
  "digest",
  "hex-literal",
- "ppv-lite86",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


newpavlov:
Closing as a clearly LLM-generated PR. The code has a number of issues and I do not want to spent my time on guiding LLM to fix them.

Issue #909, opened 2026-08-18
sha1-checked: hardware acceleration on arm64


srijs:
Hi everyone!

For a project I'm currently working with `sha1-checked` and have integrated the hardware acceleration available on `arm64` for about ~2.2-2.4x better performance compared to the baseline, and landing within ~25% of plain unchecked SHA-1.

 The downside is that it needs a little bit of `unsafe` code around the `core::arch::aarch64` instrinsics, as well as a bit more complexity due to the runtime feature detection and branching.

Would there be interest in me upstreaming these changes?

newpavlov:
Could you share a repo with current code?

>The downside is that it needs a little bit of unsafe code around the core::arch::aarch64 instrinsics, as well as a bit more complexity due to the runtime feature detection and branching.

Use of intrinsics themselves is safe on functions marked with appropriate `#[target_feature(enable = "..")]`, we would still need `unsafe` to call such function, but it should be just one line. Though we would need to slightly bump MSRV for this.

>Would there be interest in me upstreaming these changes?

Yes, but I think it's better to do after the new version is released (see #908).

srijs:
Thanks for the response, @newpavlov!

Certainly happy to share the code, just wanted to see if there’s interest at all before putting in the work to clean up the code.

If you want me to bump `msrv` part of the change I’m happy to, alternatively what I usually do with `crc32fast` (which I maintain) is to do `rustc` version checks in `build.rs` and trigger a cfg feature based on that. Here’s an [example](https://github.com/srijs/rust-crc32fast/blob/16937a2e1cf880b476ca4484f52be7aa1d73b559/build.rs).

I’ll go and prep a draft PR, will circle back shortly.

srijs:
@newpavlov please find the PR here: https://github.com/RustCrypto/hashes/pull/910.

srijs:
@newpavlov is there a timeline yet for the 0.11 release?

PR #910, opened 2026-08-22
sha1-checked: implement hardware acceleration on arm64 and x86_64

srijs:
re #909

Implements hardware acceleration for `sha1-checked` on both `arm64` and `x86_64` architectures.

This is largely based on the existing `sha1` implementation, with modifications to spill W0..W79 into a buffer as the message is being expended. The way this then works is by running the happy path via the specialized hardware instructions, and falling back to the scalar implementation only if a potential collision is detected.

Also tried to match using `cpufeatures` and `cfg-if` as best as possible, yet slightly different as it's somewhat of a different surface here. Let me know where you'd like me to adjust this.

Code duplication between `sha` and `sha1-checked` is obviously one of the not so nice aspects here - if you have ideas or a desire to share the instrinsics code across the crates let me know!

---

Measured with the crate's own bench suite:

|                                                | before    | after     | speedup   |
| ---------------------------------------------- | --------- | --------- | --------- |
| **aarch64** (dedicated Apple Silicon)          | 928 MB/s  | 1996 MB/s | **2.15x** |
| **x86_64** (AMD EPYC, `sha_ni`, cloud sandbox) | ~300 MB/s | ~640 MB/s | **~2.1x** |

For comparison, performance relative to plain `sha1` on the same backend, i.e. what fraction of un-checked SHA-1's speed detection reaches:

|                       | scalar | hardware-accelerated |
| --------------------- | ------ | -------------------- |
| **aarch64**           | 64%    | 63%                  |
| **x86_64** (`sha_ni`) | ~55%   | ~38%                 |

_Note: On `aarch64`, hardware acceleration speeds up `sha1` and `sha1-checked` by about the same factor, so the relative gap to plain SHA-1 barely moves. On `x86_64` it doesn't, instead the gap actually *widens* (but of course still improving the absolute performance)._


sha1-checked/src/compress/soft.rs
srijs, line 1: This is existing code, moved over from `compress.rs` and with some `pub(crate)` exports added.

sha1-checked/src/compress/backend/aarch64_sha2.rs
srijs, line 1: This is more or less the same as the `sha1` [implementation](https://github.com/RustCrypto/hashes/blob/07d370c4a0d7cfab1f2a2e3d21bbb44e5669ee32/sha1/src/compress/aarch64_sha2.rs), with two key differences:

1. It's just a single block rather than a loop over blocks.
2. We spill the word sequence (W_0 .. W_79) into `w` as they are expanded across the rounds.

sha1-checked/src/compress/backend/x86_sha.rs
srijs, line 1: Same deal here, based on the original `sha1` impl, cut down to a single block, and added word spilling.

PR #915, opened 2026-09-12
sha2: use inline assembly instead of unstable intrinsics for riscv-zknh backend

TechnoPorg:
This allows the backend to be used on stable Rust.

Although it behaves in the same way as before, the codegen is slightly different now, and I haven't been able to pin down why.


newpavlov:
I don't think we should use `asm!` polyfill for this. Instead it's better to push for stabilization of the relevant intrinsics in `std`, see https://github.com/rust-lang/rust/issues/114544.

TechnoPorg:
Fair enough! I am still new to contributing to r-l/r but would like to do more, so I will start taking a look at that issue and I will see if there's anything I can to do move it forward.

PR #917, opened 2026-09-16
md5: add x86_64 and AArch64 asm backends

sjthomason:
Port the implementations to Rust inline assembly based on: https://github.com/animetosho/md5-optimisation

The original work was placed in the public domain: https://github.com/animetosho/md5-optimisation/discussions/4


md5/src/compress/x86_64_asm.rs
```diff
@@ -0,0 +1,864 @@
...
+#[allow(clippy::too_many_lines)] // intentionally monolithic for the assembler
+#[allow(clippy::many_single_char_names)] // a/b/c/d/m are standard MD5 register names
+#[allow(clippy::cast_ptr_alignment)] // read_unaligned semantics via asm memory operands
+#[allow(clippy::cast_possible_wrap)] // K constants are bit-patterns; sign is irrelevant
+#[inline(always)]
+fn compress_block(state: &mut [u32; 4], block: &[u8; 64]) {
+    // SAFETY: all pointer arithmetic stays within the 64-byte block and the
+    // 4-word state array.  Register constraints satisfy Rust inline-asm rules.
+    unsafe {
+        use core::arch::asm;
```
oech3: `use` can be outside of `unsafe {}`.

md5/src/compress/x86_64_asm.rs
```diff
@@ -0,0 +1,864 @@
...
+
+        let mut a = state[0];
+        let mut b = state[1];
+        let mut c = state[2];
+        let mut d = state[3];
+        let (sa, sb, sc, sd) = (a, b, c, d);
+
+        // Scratch registers shared across both asm blocks.
+        let mut t1: u32; // rolling C copy / xor-and scratch
+        let mut t2: u32; // G second-term scratch
```
oech3: should be outside of `unsafe {}`.

oech3:
> The original work was placed in the public domain: https://github.com/animetosho/md5-optimisation/discussions/4

For maintainers, the original work changed license to Apache2.0/MIT .

tarcieri:
I think it'd be helpful if this were split up into separate PRs for AArch64 vs x86(_64). Even #447 went by the wayside because these kinds of ASM backends are hard-to-review, even when adapted from seemingly mature sources of ASM. Though while this ASM seems to have a decent rationale, I'm not finding a whole lot in the way of a test suite.

Proptests which compare the Rust reference implementation with the ASM implementation would be nice. Also from what I can tell CI isn't even testing `aarch64` whatsoever, just ensuring it compiles.

It would also be helpful if these changes were justified with benchmarks.

Edit: I looked for formally verified ASM for MD5 and wasn't able to find any

oech3:
> I looked for formally verified ASM for MD5 and wasn't able to find any

Author of https://github.com/animetosho/md5-optimisation said most of them are included to OpenSSL. It might be able to reference OpenSSL's md5.

